PR #8, opened 2020-03-11
DDA-7: Channels and data models

jvalanen:


docs/channels.yaml
```diff
@@ -0,0 +1,40 @@
...
+                    type: "object"
+                    schema:
+                      $ref: "#/definitions/diory"
+        error:
+          description: "Error response"
+          schema:
+            type: "string"
+            description: "Error message"
+definitions:
+  diory:
```
jvalanen, line 31: Should also room and home have own definitions?

docs/data-models.md
```diff
@@ -0,0 +1,30 @@
...
+```
+id: uri
+rooms:
+    [path]: diory
+```
+Diory is in the room diograph with room id
+
+**Room**
+```
+id: uri
```
jvalanen, line 27: Should this uri be room's path? The same as in home's room array?

electron/channels/get-room-channel.js
```diff
@@ -0,0 +1,10 @@
+const { ipcMain } = require('electron')
+const { channels } = require('../../src/shared/constants')
```
jvalanen, line 2: I'm thinking if channelNames or channelIds is better... This sounds like it's the channel itself. But of course when you look at /shared/constants you understand that it's not... If you like it this way, it's ok for me.

electron/main.js
```diff
@@ -1,5 +1,6 @@
 const { app, BrowserWindow, ipcMain } = require('electron');
 const path = require('path');
+const glob = require('glob');
 const url = require('url');
 const fs = require('fs');
 const { channels } = require('../src/shared/constants');
```
jvalanen: There's now unnecessary imports here

jvalanen:
Very important refactoring and documentation, to get good base!

Documentation can be pretty simple now. We can go into details later when things stabilize.

public/home.json
```diff
@@ -2,12 +2,12 @@
...
   "rooms": {
-    "room1": {
+    "./public/room1": {
       "id": "room1",
       "text": "Room 1",
       "image": "https://pix6.agoda.net/hotelImages/4656079/-1/f7771c6afc7cc32401286116a7eed6f0.jpg?s=1024x768"
     },
-    "room2": {
+    "./public/room2": {
       "id": "room2",
```
jvalanen: So this is the id which identifies this room when it's shared to public/other people?

Should it be a bit more "id like", if not uuid, maybe something like: "room2-it2je7idkd3"

docs/channels.yaml
opohjola: yes, if we think that this is is a good format. Is it?

docs/data-models.md
opohjola: it's room id, not path

electron/channels/get-room-channel.js
opohjola: this was used in template but I agree that they are misleading.

opohjola: used also in ipcRenderer, so let's leave it

docs/channels.yaml
jvalanen: I don't see any problem with this or "know about anything better" :)

docs/data-models.md
jvalanen: Yeah, I understand this better now.

But in Home definition there is still `[path]`? Is that still correct? Or should it be `[id]`?

docs/data-models.md
```diff
@@ -0,0 +1,30 @@
...
+data: object
+links: 
+    [path]: diory
+```
+
+**Home**
+```
+id: uri
+rooms:
+    [path]: diory
```
jvalanen: Should this be `[id]` if the connector is the one responsible of paths?

public/home.json
opohjola: yes, it could be. Let's fix this in DDA-5

PR #12, opened 2020-03-14
DDA-11: Connector Feature

opohjola:
Refactor project structure
Create Connector feature to store paths to rooms
Move respective functions to connector feature


electron/main.js
```diff
@@ -1,5 +1,6 @@
 const { app, BrowserWindow, ipcMain } = require('electron');
 const path = require('path');
+const glob = require('glob');
 const url = require('url');
 const fs = require('fs');
 const { channels } = require('../src/shared/constants');
```
jvalanen, line 6: Unnecessary imports (same comment as in other pull request)

src/react/features/connector/connectors.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Channel
+
+  Background:
+    Given I home
```
jvalanen, line 4: Hmm? `Given I'm in home`?

jvalanen: Should these be tested as features? I think these are not "user stories"... 🤔 

I propose deleting or completing this .

src/react/features/connector/hooks/index.js
```diff
@@ -0,0 +1,13 @@
+import { useGetHome } from './useGetHome'
+import { useGetRoom } from './useGetRoom'
+import { useSaveHome } from './useSaveHome'
+import { useSaveRoom } from './useSaveRoom'
+
+export { useAddRoom } from './useAddRoom'
+
+export const useChannel = () => {
+  useGetHome()
+  useSaveHome()
```
jvalanen, line 10: I wonder why we need to "save home" at startup? 

"save room" we need because room is "created" by reading the folder if diograph.json is not available but isn't home always just home.json. So at startup we just read the home.json and save the same home.json again?

jvalanen:
What's the difference between GET_ROOM.json and diograph.json? And same way GET_HOME.json and home.json?

Is GET_ROOM.json or GET_HOME.json used anywhere? Or are they just old obsolete stuff?

jvalanen (approved):
I read this through. Couldn't digest everything at first sight but learned and understood a lot.

Added few comments but nothing too significant. Check also my comments from the other pull request (https://github.com/DioryMe/diory-browser-electron/pull/8) as these are combined.

There's still lots of going on and it takes time that it stabilizes (and along that I'll also understand more...)

src/react/features/connector/connectors.feature
opohjola: Yes, that might be. I'll remove this for time being.

src/react/features/connector/hooks/index.js
opohjola: It says useSaveHome, not save home :) Basically, it initializes the subscription to the home and room changes that triggers the the actual save action.

opohjola:
> What's the difference between GET_ROOM.json and diograph.json? And same way GET_HOME.json and home.json?
> 
> Is GET_ROOM.json or GET_HOME.json used anywhere? Or are they just old obsolete stuff?

Idea is to use channel names in mock responses. Fixed that. Those are used in "fileClient" and feature tests.

opohjola:
Noticed that deleting base branch closes the PR and then basebranch cannot be changed. Rebased manually to develop.

PR #14, opened 2020-03-30
Github Actions + initial docker development environment

jvalanen:


docs/development-environment.md
```diff
@@ -0,0 +1,33 @@
+# Development environment
+
+## Known issues
+
+* [ ] When running e.g. `yarn upgrade --latest` in container, yarn.lock is not modified
+* [ ] Husky is not working in container, so local yarn install is required
```
opohjola: Is husky just for prettier? Do we get the same result (i.e. running prettier automatically) with combining prettier to eslint? prettier-eslint -> eslint --fix?

opohjola:
I have to understand more to give more comments. Good to go!

opohjola (approved):

docs/development-environment.md
jvalanen: Currently it is set in `package.json` like this:
```  
  "husky": {
    "hooks": {
      "pre-commit": "lint-staged"
    }
  },
  "lint-staged": {
    "{electron,src}/**/*.{js,jsx,ts,tsx,json,css,scss,md}": [
      "prettier --write",
      "git add"
    ]
  }
```

Haven't touched that. 

But you're right! We could run also `eslint --fix` the same way as prettier is run now and that will fix automatically also eslint stuff. Hadn't thought about that but I think it's ok. 

But I won't broaden the scope here. Let's do it in some future PR.

jvalanen: Husky had started to work in container! So I removed those todo things and try out making commit inside of the container...let's see if it's too slow though :/ But I want to test if it's possible to run our development environment without local Node version hassle or `yarn install`.

Issue #6, opened 2020-03-09
DDA-6: Generate diory links from subfolders


opohjola:
Given I have a folder with subfolders
When I take folder diory in focus
Then I see subfolder diorys in view

opohjola:
#16

PR #16, opened 2020-04-02
DDA-6: Generate diograph from folders and subfolders

opohjola:
#6


docs/react.md
```diff
@@ -23,7 +23,7 @@ Test higher in the tree
 - types -> typescript --checkJS (?)
 - format -> prettier
 - logic -> jest
-- feature -> cucumber, jsdom (spectron?)
+- feature -> cypress.io, spectron?, cucumber, jsdom
```
jvalanen, line 26: Do you already have something cypress.io stuff here?

docs/read-folder-tool.md
```diff
@@ -11,3 +11,31 @@ Compose diograph.json from folder contents
...
+
+
+## Open questions
+
+### Diory
+### Id
+- id can be anything
+- uuid
+- content hash -> ipfs -> new versions using ipns
+- created time, great for timeline
```
jvalanen, line 23: Do you mean created time as an id? I think "createdTime" attribute would be better for created time. And if timestamp is used at id, I'd still suggest having some random characters in the end to ensure the uniqueness.

electron/main.js
```diff
@@ -14,24 +15,14 @@ function createWindow() {
     },
   })
 
-  if (process.env.DEV_BUILD) {
-    // Enable loadFile: showing images from external folder
-    // - disables using localhost:3300 & React hot-reload
-    // - requires running `yarn run build`
-    mainWindow.loadFile('build/index.html')
```
jvalanen, line 21: Is this now removed? Or moved? Or deleted by accident?

electron/main.js
```diff
@@ -2,6 +2,7 @@ const { app, BrowserWindow } = require('electron')
 const path = require('path')
 const glob = require('glob')
 const url = require('url')
+require('electron-reload')
```
jvalanen, line 5: This sounds good! Did it work?

package.json
```diff
@@ -61,6 +66,8 @@
     "eslint": "eslint ./{src,electron}/. --ext .js",
     "prettier": "prettier --write '{electron,src}/**/*.{js,jsx,json,css,scss,md}'",
     "test": "react-scripts test",
+    "test-react": "react-scripts test",
+    "test-electron": "jest --detectOpenHandles electron/**/*.js --watch --verbose",
```
jvalanen, line 68: Oh, doesn't react-scripts test run also jest tests? I prefer having one command which runs all the tests.

jvalanen: Aah, now I understand, it's that --watch! And that `yarn test` is run in Github Actions with env `CI=true` which excludes watch.

package.json
```diff
@@ -95,6 +102,9 @@
     "presets": [
       "@babel/preset-env",
       "@babel/preset-react"
+    ],
+    "plugins": [
+      "@babel/plugin-transform-runtime"
```
jvalanen, line 105: Can you describe what is this?

electron/generators/diograph-generator.js
```diff
@@ -0,0 +1,68 @@
+const { isEmpty, promiseAllReduce } = require('../lib/utils')
```
jvalanen, line 1: Are these generators only part of the "read folder tool" or are they used elsewhere?

Should we already have a concept of "tool" at least in the code-folder level? That all the code related to one tool would be in their own folder. In this case `electron/tools/read-folder/generators/diograph-generator`

jvalanen (commented):
These were my first thoughts and questions :) Seems like a deep dive and a whole lot more changes than I expected. But it's very important to get our dev-flow conventions in place, especially testing habits!

Some kind of conceptual/architectural description about the generators (diory, link, diograph) and readers (file, folder, image). Kind of understand the big picture a bit better and then make another review.

Maybe extend the `read-folder-tool.md` with these new concepts?  (e.g. what kind of new generators or readers we may need in the future?)

electron/channels/add-room-channel.js
```diff
@@ -1,14 +1,20 @@
...
+    console.log(id, diograph)
     console.log('--------')
-    event.reply(channels.ADD_ROOM, room)
+    saveRoom(path, diograph)
+      .then(() => {
+        event.reply(channels.ADD_ROOM, diograph[id])
+      })
+      .catch(err => {
+        console.log(err)
+        return event.sender.send(channels.ADD_ROOM, null, err)
```
jvalanen, line 17: That's great that we have exception handling adn this looks good to me!

Could you make a new error-handling.md or channels.md (or somewhere) and describe what is our convention to handle errors with ipcRenderer channels?

How should we show errors to users in UI? Alert box, notification?

electron/main.js
opohjola: no, there is a separate script for that: "build-and-start-electron"

electron/main.js
opohjola: jep

package.json
opohjola: there could be on script (test) that runs all tests, but for development it helps to have separated ones with appropriate options.

electron/generators/diograph-generator.js
```diff
@@ -0,0 +1,68 @@
...
+  return {
+    id: diory.id,
+    diograph: {
+      ...diograph,
+      ...addToDiograph(diory),
+    },
+  }
+}
+
+exports.generateDiograph = generateDiograph
```
jvalanen, line 70: So generateDiograph is the only "public interface" of the read folder tool?
Sounds good to me, very simple :)

Could this file be called read-folder-tool.js? Or path be `electron/tools/read-folder/index.js`?

Because isn't this what we call "Read folder tool"?

package.json
opohjola: it is needed for jest to run tests in node.

electron/generators/diograph-generator.js
opohjola: maybe. Let's see what happens when the next tool comes.

electron/generators/diograph-generator.js
```diff
@@ -0,0 +1,68 @@
...
+    ...foldersDiograph,
+  }
+
+  const links = {
+    ...folderLinks,
+    ...fileLinks,
+  }
+
+  const linkedDiorys = getDiorys(diograph, links)
+  const diory = {
```
jvalanen, line 56: Is this the "room diory"? Could the const be named `const roomDiory`? Or room? 🤔

electron/generators/diograph-generator.js
```diff
@@ -0,0 +1,68 @@
+const { isEmpty, promiseAllReduce } = require('../lib/utils')
+const { readPaths } = require('../readers/folder-reader')
+const { generateFileDiory, generateFolderDiory } = require('./diory-generator')
+const { generateFileLink, generateFolderLink } = require('./link-generator')
+
+function getDiorys(diograph, links) {
```
jvalanen, line 6: I'd prefer having public interface first and these helper functions later.

My order for these functions would be:
* generateDiograph
* generateFileDiographAndLink
* generateFolderDiographAndLink
* getDiorys
* addToDiograph

This is just my "feeling" and I know some teams prefer something like alphabetical order (which is objective and either right or wrong, no opinionated) but my thinking is:
* public functions (in order of importance 😉 )
* 1st private function called in 1st public function
* 2nd private function called in 1st public function
* 1st function called in 2nd public function
etc.

Or most important private function first... 🤔 
Fuzzy...I know...but I like it that way! 😄

electron/generators/diograph-generator.js
```diff
@@ -0,0 +1,68 @@
...
+function addToDiograph(diory) {
+  return (
+    diory.id && {
+      [diory.id]: diory,
+    }
+  )
+}
+
+async function generateFileDiographAndLink(filePath) {
+  const diory = generateFileDiory(filePath)
```
jvalanen: Could this be `const fileDiory` instead of just generic `const diory`?

electron/generators/diograph-generator.js
jvalanen, line 70: Elsewhere you haven't used `exports.` on its own line but as a part of the function definition. Is there a reason that you did this way in here or is this incorrect and should be fixed?

Elsewhere it's like:
```
exports.generateFileDiory = function generateFileDiory(filePath) {
...
}
```

electron/generators/diograph-generator.spec.js
```diff
@@ -0,0 +1,164 @@
+const { readPaths } = require('../readers/folder-reader')
+const { generateFileDiory, generateFolderDiory } = require('./diory-generator')
+const { generateDiograph } = require('./diograph-generator')
+
+jest.mock('../readers/folder-reader')
+jest.mock('./diory-generator')
+
+describe('diograph-generator', () => {
```
jvalanen, line 8: I'd prefer having the function name instead of file name here: `generateDiograph`
```suggestion
describe('generateDiograph', () => {
```

electron/generators/diograph-generator.spec.js
```diff
@@ -0,0 +1,164 @@
+const { readPaths } = require('../readers/folder-reader')
```
jvalanen, line 1: Good job Oopee! You have correctly tested only public interface and mocked and verfied the called values exactly the same way that I'd have done 😊 . And very profound testing with so many different combinations! 🎉  (I wouldn't have done that many because my laziness....)

jvalanen (commented):
2nd round done, looks good!

I'd take a look one more time once you reply to the comments / get the changes done.

docs/react.md
opohjola: not yet

docs/read-folder-tool.md
opohjola: we need an id that somehow links to the actual file because if and when we read the folder again all manual links should remain. This can be done using a content hash. I was thinking that created time would be easier and simpler but probability for same ids is too high.

electron/channels/add-room-channel.js
opohjola: Let's have a separate issue to backlog on this. I'm not very confident on this.

opohjola: I added a depth section along bugs. Not sure if it's the right place.

electron/generators/diograph-generator.js
opohjola: Not sure yet; how channels link to tools? Is ADD_ROOM_CHANNEL part of read-folder-tool? What is the react end of the read-folder-tool?

opohjola: At the moment I do not call it 'read folder tool' :) I don't have a name for it. So I propose to wait a bit; thing will clarify.

electron/generators/diograph-generator.js
opohjola: No, it's diory of each folder and subfolder. Maybe the ADD_ROOM_CHANNEL is not correct; maybe it's just GENERATE_DIOGRAPH_CHANNEL...

electron/generators/diograph-generator.js
opohjola: I think (and eslint confirms) that functions should be introduced before their usage. I prefer the order they are used. And thus public functions in the bottom :)

If we would use const -notation for function, this would be required by javascript.

opohjola: 28:34  warning  'generateDiograph' was used before it was defined  no-use-before-define

This don't have other option as it's recursive.

opohjola: One option is the place private functions inside public but I think it makes readability harder.

electron/generators/diograph-generator.js
opohjola: true, it's in consistent. I'll prefer end so I would change all others like:
exports.generateFileDiory = generateFileDiory
exports.generateFolderDiory = generateFolderDiory
but this relates to discussion on order of functions

electron/generators/diograph-generator.spec.js
opohjola: ah, I usually have both.

electron/generators/diograph-generator.spec.js
opohjola: Well, I would have liked to test many modules at once but it caused problems and I wanted to spec the behavior better so that's why I choosed to mock most of submodules.

opohjola: So, I wouldn's say this is the only correct way to do tests/specs.

electron/generators/diograph-generator.js
opohjola: I wont's change now.

opohjola:
I would not document the internal stuff; the code should speak for itself and if it does not, it should be improved. However, some list of implemented, missing and foreseen features could be added along the open issues that was already started.

docs/read-folder-tool.md
jvalanen: Aah, now I understand, that's true. Content hash sounds good to me. And I think that's the only viable option that I can think of...

electron/channels/add-room-channel.js
jvalanen: 👌

electron/generators/__snapshots__/diograph-generator.snapshot.test.js.snap
```diff
@@ -0,0 +1,70 @@
...
+    "image": "/Users/op/2019/diory-browser-electron/electron/readers/example-folder/other-image.jpg",
+    "latitude": 43.464455,
+    "longitude": 11.881478333333334,
+    "modified": "2020-04-02T17:57:14.569Z",
+  },
+  "uuid-1": Object {
+    "created": "2020-04-02T17:57:14.572Z",
+    "date": "2019-11-25T19:58:24.000Z",
+    "id": "uuid-1",
+    "image": "/Users/op/2019/diory-browser-electron/electron/readers/example-folder/some-image.jpg",
```
jvalanen, line 18: This is why we need docker environment 😄  As you can imagine, this test won't pass on my computer...

But this is a good starting point, we can fix this later...

electron/generators/diograph-generator.js
jvalanen: Okay, interesting... Have to think about this...let's get back to this in next pull requests :)

electron/generators/diograph-generator.js
jvalanen: Aah, ok, true.

Interesting: ADD_ROOM_CHANNEL vs. GENERATE_DIOGRAPH_CHANNEL

electron/generators/diograph-generator.js
jvalanen: Okay, let's wait and see :)

jvalanen: Okay, this is not a big thing but what is a big thing is "consistency" 😄

electron/generators/diograph-generator.spec.js
jvalanen: Aah, you have also the `generateDiograph` there!

electron/main.js
jvalanen: Okay... If you said you tested this with your own content, then it must work! Have to try out myself too! :)

Realised also that the pathname had changed a bit:
```
// New
pathname: path.join(__dirname, '../build/index.html'),
// Old
pathname: path.join(__dirname, '../index.html'),
```

jvalanen:
I think we are good to go 😊 💪

PR #18, opened 2020-04-09
Yarn upgrade & Prettier 2.0 changes

jvalanen:
* Rebased on #14  
* `yarn upgrade --latest` 
* `yarn run prettier` which applied new defaults of the Prettier 2
* Fixed Security Alerts that Github was complaining about. Needed to add resolutions attribute to `package.json` and run `yarn install`
* Sketched an initial "How to upgrade packages?" instructions to `docs/upgrade-packages.md`


docs/upgrade-packages.md
```diff
@@ -0,0 +1,62 @@
...
+1. Yarn is updated as a part of the Docker image.
+  * Docker image's Dockerfile and YARN_VERSION from [Docker hub](https://hub.docker.com/_/node/) by selecting the version
+
+**NOTE:** After this kind of update everybody should rebuild their docker images (`docker-compose build`) and/or set Node to the specific version on their local machine (e.g. `nvm use 12.13`) (and maybe also somehow update yarn?)
+
+## FAQ
+
+### How often this should be done?
+
+At least once a quarter OR when major
```
opohjola, line 47: Can we have some notification of these needs to upgrade so that we remember?

opohjola (approved):

docs/upgrade-packages.md
```diff
@@ -0,0 +1,62 @@
...
+**NOTE:** After this kind of update everybody should rebuild their docker images (`docker-compose build`) and/or run `yarn install` on their local machine.
+
+## Security fixes
+
+Sometimes `yarn upgrade --latest` doesn't fix all the security issues.
+
+Then it's ok to use add the patched version of the package in the [resolutions in package.json](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/):
+
+```js
+"resolutions": {
```
opohjola, line 23: So these packages are something else than the latest?

opohjola (approved):
I was thinking that could/should we have the DoD list in the PR: e.g.

[x] unit tests
[-] feature tests 
[x] documentation 
[-] storybook
...
x -> done
- -> not applicable
or something...

Or is this too much overhead?

package.json
```diff
@@ -95,14 +105,17 @@
     "presets": [
       "@babel/preset-env",
       "@babel/preset-react"
+    ],
+    "plugins": [
+      "@babel/plugin-transform-runtime"
     ]
   },
   "prettier": {
     "trailingComma": "es5",
```
opohjola: I understood that this is now the default and can be removed.

opohjola (approved):

docs/upgrade-packages.md
jvalanen: Yes. These are dependencies of our project dependencies. `yarn upgrade --latest` upgrades all *our* dependencies to the latest but not all the sub-dependencies to the latest. When e.g. `react-scripts` is updated then we don't need to list `acorn` in here anymore because it will be resolved also without this kind of "forcing".

This was just something I "read from the internet" and understood that is the thing to do in this kind of situations 😊 (see link above^ https://classic.yarnpkg.com/en/docs/selective-version-resolutions)

docs/upgrade-packages.md
jvalanen: I don't think we need to. It's not that serious. And anyway few times a year for some reason we need to deal with library versions and then if it's very long from the previous upgrade, we can create a separate ticket to address the need.

jvalanen:
> I was thinking that could/should we have the DoD list in the PR: e.g.
> Or is this too much overhead?

I think it's a good idea but maybe not yet. Let's let this DoD thing marinate a bit and then see if we want this. 

I suggest that we go through the "working draft DoD" informally and try to put it that way in practice little by little...

package.json
jvalanen: Yes, I'll remove this.

PR #19, opened 2020-04-09
DDA-17: Add room to home

opohjola:
[x] refactor
[x] linter
[ ] test coverage
[ ] document
[x] coding style
[ ] feature tests
[ ] storybook
[ ] unit tests
[x] manual testing
[x] changelog


src/react/features/connector/hooks/useSaveHome.js
```diff
@@ -1,10 +1,10 @@
...
 import { useStore, usePromiseDispatch } from '../../../store'
 
-import { saveRoom } from '../../room/actions'
-
-import { connect } from '../client'
-import { channels } from '../../../../shared/constants'
+// import { saveRoom } from '../../room/actions'
+//
+// import { connect } from '../client'
+// import { channels } from '../../../../shared/constants'
```
jvalanen: Why these weren't removed but just commented out?

I suggest removing rather than commenting out.

package.json
```diff
@@ -1,21 +1,21 @@
...
   "name": "diory-browser-electron",
-  "version": "0.3.3",
+  "version": "0.3.4",
   "productName": "Diory - Digital Memory Browser",
   "main": "electron/main.js",
   "private": true,
   "dependencies": {
     "evergreen-ui": "^4.24.0",
-    "exifreader": "^3.6.0",
+    "exifreader": "^2.13.0",
```
jvalanen, line 9: Can you describe why you downgraded and document it somewhere so we remember 
1) not try to upgrade this package next time (and get the same error)
2) why we shouldn't upgrade this package
3) what we should do before we could upgrade this package

Or is it too much? Or do you have other suggestions?

package.json
```diff
@@ -1,21 +1,21 @@
...
     "glob": "^7.1.6",
     "leaflet": "^1.6.0",
     "lodash": "^4.17.15",
     "prop-types": "^15.7.2",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
-    "react-scripts": "3.4.1",
+    "react-scripts": "3.2.0",
```
jvalanen, line 16: Same thing for this than with the `exifreader`

package.json
```diff
@@ -1,21 +1,21 @@
...
     "ui-box": "^3.2.0",
-    "uuid": "^7.0.3"
+    "uuid": "^3.3.3"
```
jvalanen, line 18: Same thing for this than with the `exifreader`

jvalanen (approved):
My comments: :D

1) There's so many changes and refactoring etc. that I don't understand enought to comment :( I read it through but feels like hebrew to me...

2) I maybe should take some "modern react" course to understand those `useSomethingRoom`, `useEffect`, dispatch, reduce, hooks and actions... :/ Any suggestions? 
Which React version the hooks came and change it all?

3) If you merge this and I start my own ticket based on this, I may have some more comments when I get the UI running etc.

4) Keep on going if you feel confident, I don't want to hinder you. But after the first release we should maybe:
1) go through together the whole code base (so I also understand...)
2) document the basic interfaces and try to stabilize them

That's my five cents, all I've got 😊 🙏

package.json
opohjola: I added these to debt.

opohjola: apparently except exifreader, which I added now

src/react/features/connector/hooks/useSaveHome.js
opohjola: True, the whole useSaveHome is a placeholder. I'll remove it.

PR #26, opened 2020-05-17
DDA-25: Remove home & store roomInFocus with electron-store

jvalanen:
* Use electron-store (https://github.com/sindresorhus/electron-store) to store room info
* GET_ROOM returns room with a stored roomId
* Remove GET_HOME channel & hook
* Store roomId on GENERATE_DIOGRAPH
* uuid package upgraded to 7.x


README.md
```diff
@@ -3,19 +3,22 @@ This project was bootstrapped with [Create React App](https://github.com/faceboo
...
 1. `yarn start` (in one terminal tab)
 1. `yarn start-electron-dev-tools` (in another terminal tab)
 1. Electron app is launched and loads `http://localhost:3300` to its window
 
-### Local development with access to folders
-1. Install Node 12 and yarn
-1. `yarn install`
-1. `yarn start-electron-dev-build`
-1. Electron app is launched and loads `./build` folder to its window
-
```
jvalanen: This should be kept?

src/react/features/connector/hooks/useGetRoom.js
```diff
@@ -1,22 +1,30 @@
...
-
   const dispatch = useDispatch()
   useEffect(() => {
-    if (roomId) {
-      connect(channels.GET_ROOM, path).then(({ diograph }) =>
-        dispatch(getRoom({ id: roomId, diograph }))
-      )
-    }
-  }, [roomId, path, dispatch])
+    connect(channels.GET_ROOM).then(({ roomId, path, diograph, dioryIdInFocus }) => {
```
opohjola: There are couple of conceptual problems here for me:
1. In case of one copy of a room, path to room is ok, but once shared, it's irrelevant for others what is my path to the room. Thus path could be e.g. a property of connector to the room, not the room itself. 
2. I think dioryInFocus is a property of app, not room. I could be a property of room but then we should update the focus everytime the focus is changed.

opohjola: Thus in my opinion, the connector should be responsible of storing its path to room to backend. 
Here, we could listen to path changes -> getRoom from path -> dispatch getRoom and enterRoom

opohjola: Note that enterRoom already has the logic to set the "room diory" into focus.

electron/channels/get-room-channel.js
```diff
@@ -1,12 +1,25 @@
...
-    console.log(JSON.stringify(room, null, 2))
+  const store = new UserDataStore()
+
+  const defaultRoom = {
+    roomId: 'welcome-room',
+    path: './public/default-welcome-room',
+    dioryIdInFocus: 'welcome-room',
+  }
+
+  const roomInFocus = store.get('roomInFocus') || defaultRoom
```
opohjola, line 14: I would store just path in GET_PATH/SET_PATH or similar (see more below)

electron/channels/add-room-channel.js
```diff
@@ -1,20 +0,0 @@
-const { ipcMain } = require('electron')
```
opohjola, line 1: I was using "DELETE:" tag :)

electron/channels/generate-diograph-channel.js
```diff
@@ -1,26 +1,26 @@
...
-  generateDiograph(path).then(({ id, diograph }) => {
-    console.log(id, diograph)
+  generateDiograph(path).then(({ roomId, diograph }) => {
+    console.log(roomId, diograph)
     console.log('--------')
     saveRoom(path, diograph)
       .then(() => {
+        // Store current roomId to user data
         const store = new UserDataStore()
         store.set({
```
opohjola, line 18: I would keep every channel simple and the most of the logic in the front-end. Thus we could have a separate channel for storing necessary information to backend.

src/react/features/connector/tools/useOpenFolder.js
```diff
@@ -21,13 +22,15 @@ export const useOpenFolder = () => {
       dispatch(setInactive())
       window.nativeFileDialog.showOpenDialog({ properties: ['openDirectory'] }).then((result) => {
         const path = result.filePaths[0]
-        connect(channels.GENERATE_DIOGRAPH, path).then(({ id }) => {
-          console.log('Open folder')
```
opohjola: I would like to have this kind of set of logging in all channels.

src/react/features/connector/hooks/useGetRoom.js
opohjola: In path storing, connector feature should be changed to store just one path, it should also GET_PATH initially from backend and SET_PATH to backend in openFolder.

opohjola (commented):
I propose what we store only path to folder in backend and that is responsibility of the folder connector, as well as it's retrieval in startup.

electron/channels/generate-diograph-channel.js
jvalanen: Is there any other logic in backend than individual channels?
* GENERATE_DIOGRAPH
* GET_DIOGRAPH / GET_ROOM
* SAVE_DIOGRAPH / SAVE_ROOM
* GET_PATH / GET_USER_DATA
* SET_PATH / SET_USER_DATA

Should we treat the backend like API? Organise everything based on channels? Channels would be kind of API endpoints 🤔 And backend is nothing else than an "channel API". Then we could also make a "API docs" for it: request parameters, response example etc.

Other endpoints would get data from folders, others from user data... But for frontend it wouldn't make any difference.

electron/channels/add-room-channel.js
jvalanen: Yeah, this deleting should be done the same way as you delete the home stuff...

src/react/features/connector/hooks/useGetRoom.js
jvalanen: This won't be done by now. Let's consider and change the architecture later...

electron/channels/add-room-channel.js
jvalanen: Commit message change done.

electron/channels/generate-diograph-channel.js
jvalanen: This won't be done by now. Let's consider and change the architecture later...

Also the naming will be reconsidered later.

src/react/features/connector/tools/useOpenFolder.js
jvalanen: Logging is again there!

electron/channels/get-room-channel.js
jvalanen: This won't be done by now. Let's consider and change the architecture later...

opohjola:
As discussed, let's return to this when refactor the architecture after the next release. In proposed architecture, connector stores the latest path and gets diograph from the path.

opohjola (approved):

PR #27, opened 2020-05-17
DDA-23 Clean code for release

opohjola:
[x] Remove home
[x] Remove multiple folders in room
[x] Remove map tools
[x] Own tag to make it easy to find it in the future: DELETE
[x] Use named imports
[ ] Remove creating, editing and removing actions


jvalanen (approved):
I agree with everything what is deleted here. There might be a bit more to be deleted (if you search for 'home' in the codebase) but I think we can continue deleting in my DDA-25 when GET_HOME channel gets removed etc.

Good job! Let's remove these! 😊

PR #31, opened 2020-05-23
TECH: Exclude diograph.json and .DS_Store

jvalanen:


electron/readers/dirent-reader.js
```diff
@@ -1,5 +1,8 @@
 const { resolve } = require('path')
 
+exports.isValid = function isValid(dirent) {
```
opohjola, line 3: We would need a list excluded files but let's do it later.

opohjola (approved):

PR #42, opened 2020-07-10
DDA-41: Videos to test content

jvalanen:


src/react/features/connector/mockResponses/GET_ROOM2.json
```diff
@@ -10,10 +10,30 @@
...
+      "id": "cbe32a6d-5bc3-44d0-b680-3f4a63765f73",
+      "text": "4_5807613915332347351.mp4",
+      "image": "development-content-room/4_5807613915332347351.mp4",
+      "created": "2020-07-05T09:39:20.000Z",
+      "modified": "2020-07-05T09:39:20.000Z"
+    },
+    "5456c2c3-4a69-4d80-bd2f-caa9945cff71": {
+      "id": "5456c2c3-4a69-4d80-bd2f-caa9945cff71",
+      "text": "4_5996618057181562302.MOV",
+      "image": "development-content-room/4_5996618057181562302.MOV",
```
opohjola: How about using diory.data for the video data?

src/react/features/tools/ToolView.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+import { useStore } from '../../store'
+import Tools from './index'
+
+const ToolView = () => {
+  // const [{ selectedToolId }] = useStore((state) => state.tools)
```
opohjola, line 7: This logic should be in video/hooks.js e.g. useVideo. It should include useEffect like:
```
useEffect(() => {
    function videoIsInFocus() {
       ...
    }
    if (videoIsInFocus()) {
       dispatch(setActive(videoToolId))
    }
}, [...]). 
```

and it should be initialized in useTools: useVideo().

(EDIT: Added ``` for better readability)

src/react/features/tools/ToolView.js
```diff
@@ -0,0 +1,27 @@
...
+  const diory = diograph[focus]
+
+  if (
+    diory &&
+    diory.image &&
+    (diory.image.toLowerCase().includes('.mov') ||
+      diory.image.toLowerCase().includes('.mp4') ||
+      diory.image.toLowerCase().includes('.avi'))
+  ) {
+    const selectedToolId = 'video'
```
opohjola, line 13: const [{ active: selectedToolId }] = useStore((state) => state.tools)

src/react/features/tools/ToolView.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+import { useStore } from '../../store'
+import Tools from './index'
+
+const ToolView = () => {
+  // const [{ selectedToolId }] = useStore((state) => state.tools)
+  const [{ focus }] = useStore((state) => state.navigation)
+  const [{ diograph }] = useStore((state) => state.room)
+  const diory = diograph[focus]
```
opohjola, line 10: There is a hook: useFocusDiory()

opohjola: The idea is that features provide useful hooks for other features (they are like APIs). Hmm. to be easier to find, we could add those to index.js of the feature.

src/react/features/tools/video/Video.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+
+import { useStore } from '../../../store'
+
+const Video = () => {
+  const [{ focus }] = useStore((state) => state.navigation)
+  const [{ diograph }] = useStore((state) => state.room)
+  const diory = diograph[focus]
```
opohjola, line 8: useFocusDiory()

opohjola (commented):
I propose that we use diory.data for detecting videos and VideoObject from schema.org as a data schema. 
data: 
    @context: http://schema.org
    @type: VideoObject
    contentUrl: video url

src/react/features/tools/ToolView.js
opohjola: or just use "active"

src/react/features/tools/ToolView.js
opohjola: It needs logic also for setInactive()

src/react/View.js
```diff
@@ -1,22 +1,23 @@
 import React from 'react'
```
opohjola: I would stay in "what was done and why" in commit texts.

opohjola: I would be nice to separate refactoring from the feature commits. E.g. now I could cherry-pick the refactoring commit instead of creating merge conflict by copying.

src/react/features/connector/mockResponses/GET_ROOM2.json
jvalanen: This is just PoC and I will make those changes when implementing the "video file reader" in https://github.com/DioryMe/diory-browser-electron/issues/49

src/react/features/tools/ToolView.js
jvalanen: We may have to pair-code this to get it work like this... Haven't ever done `useSomething()` hooks before :(

src/react/View.js
jvalanen: Well... There was so much going on in this commit (that I didn't even fully understand back then) that I think it's a bit too much to ask to separate something to a different commit. For me it was just "Refactoring (and some other stuff( by pair-coding with OP" :( 

But yeah, you're right. Refactorings should be done in separate commits and they should really be refactoring (="without changing its external behavior").

src/react/features/connector/mockResponses/GET_ROOM2.json
opohjola: Then it think we should clearly indicate that this is a PoC. Also do we merge PoCs into develop branch? I think not.

However, I propose that this is further developed into a production code. So this would be the front end part and #49 is the backend part. Data model would be the interface.

jvalanen: Well, this is a good discussion. Some thoughts:
1. I don't see any problem at this stage of development (=nobody uses this) to have e.g. PoC video feature in the app instead of having no videos at all. Same with everything: something that works at least somehow and doesn't break anything else is better than nothing. (in the future we should have feature-flags to control what we ship to end users)
2. I think everything should be in the same codebase/branch: PoCs, website, cloud-infra, backend, frontend, welcome content, test content... In my experience having multiple versions of code is just so much hassle and overhead that it's hard to see what would add so much value that it would be worth it (maybe there is, but I just don't see it...)
3. This codebase and our processes are so unmature/under developed yet that the most important thing is that we get forward and get code merged to the develop. We have to experiment, try and err, to find the most suitable processes for our team and get the tooling and our understanding about what we are doing and how to a level that this kind of things don't happen anymore.

With "this kind of things" I mean that I know that I'm not happy with my work in here and the way I ended up doing too many things in this PR and I completely agree with your comments and "change requests". I just want to make sure that we don't take this too seriously and the focus remains in the "getting things done" rather than "doing the things the right way" :) We can discuss this more tomorrow :)

PR #44, opened 2020-07-11
Fix package.json caused by failed rebase

jvalanen:


opohjola (approved):
Should we use same prefixes in titles as in commits? So here "FIX: ..."

PR #46, opened 2020-07-13
DDA-45: Fullscreen view

opohjola:


jvalanen (approved):
Looks good! I tried also on browser. As everything seems to work as expected I don't see reason to go into details as this will still be developed as we get more real use cases for this! And good compact PR as well!

PR #50, opened 2020-07-14
DDA-46: Set up Storybook with example components

opohjola:


docs/testing-plan.md
```diff
@@ -0,0 +1,36 @@
+# Testing plan
```
jvalanen, line 1: 1. There's also a paper about this: https://paper.dropbox.com/doc/Testing-plan--A38aCEu1lx5TW0hIVX~OLwsIAg-2NHOfYnIOmBUQxUYOYDPF
2. We should get a good routine for writing tests on all (related) levels for each task. Is this the place to describe that kind of guidelines? Or in the Development flow paper? Or in the Testing plan paper?

For me it's ok to have this kind of docs here in the `docs/` folder. And good that you wrote this!

package.json
```diff
@@ -10,15 +10,23 @@
     "exifreader": "^3.11.2",
     "glob": "^7.1.6",
     "lodash": "^4.17.15",
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1",
```
jvalanen: Why you have moved React to dependencies? Did you run into some issue?

It's not needed in dependencies as it is only used to build the the `bundle.js` and then that is used by Electron. So only dependencies required by Electron/Node needs to be in dependencies.

It's important to keep this sorting to devDependencies and dependencies in good fit because it affects to the size of our app package which is downloaded and installed by our users.

package.json
```diff
@@ -65,13 +71,15 @@
...
-    "eslint": "eslint {electron,src}/**/*.{js,jsx,json,css,scss,md}",
+    "eslint": "eslint './{src,electron}/**/*.{js,jsx,json,css,scss,md}' --fix",
     "prettier": "prettier --write '{electron,src}/**/*.{js,jsx,json,css,scss,md}'",
     "test": "react-scripts test",
     "test-react": "react-scripts test",
     "test-electron": "jest --detectOpenHandles electron/**/*.js --watch --verbose",
     "cucumber": "./node_modules/.bin/cucumber-js src/react --no-strict -r src/react/utils/cucumber --require-module @babel/register --require-module @babel/polyfill",
     "cucumber-only": "./node_modules/.bin/cucumber-js src/react/features --tags @only -r src/react/utils/cucumber --require-module @babel/register --require-module @babel/polyfill",
+    "storybook": "start-storybook -p 9009 -s public",
+    "build-storybook": "build-storybook -s public",
```
jvalanen, line 82: Do you want to host this statically somewhere? (as that what I understood is the purpose of build-storybook...)

I think it's enough to have it only in the codebase by now...

jvalanen (approved):
Good pull request! It focus only on Storybook and few components and doesn't include any "unrelated stuff".

Storybook is looks like a very good addition and great that you got it set up now! I studied a bit from the docs and think I have a good enough idea about it now. 

The React-dependency thing was only thing that may require changes but all the other comments were just questions :)

docs/testing-plan.md
opohjola: I think we could develop it in paper but then document the outcome here.

package.json
opohjola: Ah, so dependences should include only electron/node dependences. All other in devDeps? Yeah, I have to remember that this is not a fe project. I'll fix.

package.json
opohjola: No need for that sofar.

PR #74, opened 2020-12-05
DDA-67: More feature tests + UI/UX improvements

jvalanen:
Create/UpdateTool dialog UI/UX improvements:
* Focus text field when opened
* Confirm & close on enter key press
* Don't close when clicking outside of dialog (only with X, cancel or esc key)
* Latitude field before longitude field

Map/Timeline tests:
* createTool created diories appear on map/timeline
* createLocationTool created diory appear on map/timeline
* Popup appears when clicking marker and clicking popup takes diory in focus
* etc.

Other:
* Add better map-content to development content


src/react/features/lenses/map/map-create.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Creating content to Map lens
+
+  Background:
+    Given I am at home
+    When I take 'Generic content' in focus
```
opohjola, line 5: Background is context thus "And"

src/react/features/lenses/map/map-create.feature
```diff
@@ -0,0 +1,32 @@
...
+    And I take 'Diory 1' in focus
+
+  Scenario: Map with diories without locations
+    And I select map lens
+    Then I see 0 focus and 0 linked markers on map
```
opohjola, line 10: Is this a still part of background? -> Given I see 0 focus and 0 linked markers on map

src/react/features/lenses/map/map-create.feature
```diff
@@ -0,0 +1,32 @@
...
+
+  Scenario: createTool created diory appears on map (in view)
```
opohjola, line 12: I would define features in common language. e.g.
Scenario: Create new linked diory to map

src/react/features/lenses/map/map-create.feature
```diff
@@ -0,0 +1,32 @@
...
+    When I select tools button
+    And I select create button
+    And I add 'Map diory' to text field
+    And I add '62' to latitude field
+    And I add '24' to longitude field
+    And I click Done button
+    And I select map lens
+    Then I see 0 focus and 1 linked marker on map
+
+  Scenario: createTool created diory to map (in focus)
```
opohjola, line 22: Is this needed here? Focus is not a feature of create...

src/react/features/lenses/timeline/timeline.feature
```diff
@@ -2,24 +2,27 @@ Feature: Timeline lens
...
+  # - although it has a date and there's no diories in view
+  # 2. Timeline shows a marker for each diory on view
+  # - although not all of them have dates
+  # 3. Currently timeline crashes if none of the diorys in view has dates
 
-  @pending
-  Scenario: Diory popup on timeline
-    When I select diory 1
-    Then I see 1 popup on timeline
+  Scenario: Shows marker for each diory on view (although not all of them have dates)
```
opohjola, line 15: I would we nice to a "Given" for diorys without date e.g.
Given focus diory 1 have 5 linked diorys without date (check from JSON?)
And focus diory 1 have 8 linked diorys with date
Then I see 0 focus and 13 linked markers on timeline

But should it show diorys without date on timeline...?

src/react/features/tools/update/UpdateView.js
```diff
@@ -17,7 +17,11 @@ const useUpdateView = (diory = {}) => {
   const updatedDiory = { ...diory, ...values }
   const dispatch = useDispatch()
   return {
-    fields: fields.map((field) => ({ ...field, value: updatedDiory[field.key] })),
+    fields: fields.map((field) => ({
+      ...field,
+      value: updatedDiory[field.key],
+      autoFocus: field.key === 'text',
```
opohjola: Autofocus logic should be in fields.js
  {
    key: 'text',
    label: 'Text',
    format: 'string',
   autoFocus: true,
  },

src/react/features/tools/update/updateTool.feature
```diff
@@ -10,41 +10,49 @@ Feature: Update tool
...
     And I select tools button
     And I select update button
     And I take 'Diory 11' in focus
     Then I see '/test-image.png' in image field
 
   Scenario: Update image changes background image
     When I add '/test-image.png' to image field
-    And click Done button
+    And I click Done button
     Then 'diory11' has 'linear-gradient(rgba(0, 0, 0, 0.2), rgba(0, 0, 0, 0.2)), url("http://localhost:3300/test-image.png")' as 'background-image'
```
opohjola, line 50: ..includes '/test-image.png' is only relevant for the test.

opohjola (approved):
Generally ok; some improvement suggestions for tests and one change request for code in UpdateView.

src/react/features/lenses/map/map-create.feature
jvalanen: This is a mistake, should be When

src/react/features/lenses/map/map-create.feature
jvalanen: Yeah, a bit technical and common language and your suggestion is better!

But still should be exact enough e.g. if diories can be created in several ways you can't always say just "create diory"

src/react/features/lenses/map/map-create.feature
jvalanen: Well, as long as we don't have any other tests for the focus, I think it's more than ok to test it in here. If it overlaps or duplicates something then of course it should be deleted.

src/react/features/tools/update/updateTool.feature
jvalanen: I'm not sure what you meant but it has to be a real image in order to Cypress show it correctly if look the video the video about the test.

src/react/features/tools/update/fields.js
```diff
@@ -3,6 +3,7 @@ export default [
     key: 'text',
     label: 'Text',
     format: 'string',
+    autoFocus: true,
```
jvalanen, line 6: Good point, didn't realise that this is a property of the field and field properties are defined here.

And thanks for making the change for me!

src/react/features/tools/update/updateTool.feature
opohjola: I mean that we should test only relevant things e.g. 'linear-gradient...' is not. In unit tests I would write expect('background-image').toInclude('/test-image-png'), not expect...toBe('linear-gradient...

PR #79, opened 2021-01-02
DDA-000: Refactor channels and pass errors as alert dialog to frontend

jvalanen:
Backend:
- Backend IPC channels
  - Specs
  - Refactor
  - Unified logging
* JSDoc package + first documented methods as example
* Backend / production logging with electron-log => `~/Library/Logs/Diory - Digital Memory/main.log`

Frontend:
* Refactor openChannel -> invokeChannel 
* Alert to frontend
* Alternative frontend logging with electron-log => `~/Library/Logs/Diory - Digital Memory/renderer.log`


electron/channels/channel-util.js
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Wraps (1) generic IPC channel logging and (2) error handling
+ * around the eventHandler function and (3) resolves its promise
+ * and (4) replies to the IPC event from frontend
+ *
+ * @param {string} channelName - IPC channel to be used to reply to the event
+ * @param {function(): Promise} eventHandler - Promise which resolves with responseObject and rejects with errorObject
+ * @return {function(): void} Specific eventHandler function to be given for ipcMain.on as a second argument
+ */
+exports.eventHandlerWrapper = (channelName, eventHandler) => {
```
opohjola, line 16: This is somewhat different coding style than I have used to.
We might have to agree a common style so that the code will be consistent.

electron/channels/generate-diograph-channel.js
```diff
@@ -1,18 +1,33 @@
...
+ *   id: 'diograph123',
+ *   path: '/Users/...',
+ *   diograph: {
+ *      room1: { id: 'room1', image: '...', links: [...] },
+ *      room2: { id: 'room2', image: '...', links: [...] }
+ *   }
+ * }
+ *
+ */
+exports.generateDiographEventHandler = (event, path) =>
```
opohjola, line 22: We could use newer async/await and try/catch syntax.
e.g. here:
async generateDiographEventHander(event, path) {
  const { id, diograph } = await generateDiograph(path)
  await saveRoom(path, diograph)
  return { id, diograph, path }
}

opohjola: try catch is not needed here are error is passed forward as is.

electron/channels/generate-diograph-channel.spec.js
```diff
@@ -0,0 +1,44 @@
...
+const { generateDiograph } = require('../generators/diograph-generator')
+const { saveRoom } = require('../lib/room-util')
+
+// Mock event.reply
+const mockEventReply = jest.fn()
+const mockEvent = { reply: mockEventReply }
+// Mock generateDiograph
+jest.mock('../generators/diograph-generator')
+const generateDiographMock = generateDiograph.mockResolvedValue({
+  id: 'this',
```
opohjola, line 12: I propose using some syntax for generating test data.
I use typically: id: 'some-id' that says that id can be some id, not specific.
Syntax also enables using code templates, e.g., I have templates for all test blocks.
In expect, this syntax says where the data is coming from.

electron/channels/generate-diograph-channel.spec.js
```diff
@@ -0,0 +1,44 @@
...
+  diograph: 'is generated diograph',
+})
+// Mock saveRoom
+jest.mock('../lib/room-util')
+const saveRoomMock = saveRoom.mockResolvedValue(undefined)
```
opohjola, line 17: Typically there are less problems when mocks are set in beforeEach block, i.e., before each test.

electron/channels/generate-diograph-channel.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+describe('generateDiographEventHandler', () => {
+  it('sends event with generateDiograph return value', async () => {
```
opohjola, line 20: I propose writing separate test for each step, using gherkin syntax and using AAA (assign, act, assert) structure in tests.

electron/channels/get-home-channel.spec.js
```diff
@@ -0,0 +1,29 @@
...
+const { getHomeEventHandler, defaultHome } = require('./get-home-channel')
+
+// Mock event.reply
+const mockEventReply = jest.fn()
+const mockEvent = { reply: mockEventReply }
+// Mock electron-store
+jest.mock('electron-store')
+
+describe('getHomeEventHandler', () => {
+  it('empty store', async () => {
```
opohjola, line 12: gherkin syntax here:
- describe: given home store is empty,
- it: returns default home

electron/channels/get-room-channel.js
```diff
@@ -1,10 +1,21 @@
...
+ *
+ * @example Response object:
+ * { diograph: {
+ *      room1: { id: 'room1', image: '...', links: [...] },
+ *      room2: { id: 'room2', image: '...', links: [...] }
+ *   }
+ * }
+ *
+ */
+exports.getRoomEventHandler = (event, { address }) =>
```
opohjola, line 18: I favour traditional function syntax in methods instead of arrow functions (I use arrow function in map, reduce, etc.)

getRoom is already promise. No need for new Promise. My preferred format here:
function getRoomEventHandler(event, { address }) {
 return getRoom(address)
} ... I hope it works :)

electron/channels/get-room-channel.spec.js
```diff
@@ -0,0 +1,21 @@
+const fs = require('fs')
+const { eventHandlerWrapper } = require('./channel-util')
+const { getRoomEventHandler } = require('./get-room-channel')
+
+const mockEventReply = jest.fn()
+const mockEvent = { reply: mockEventReply }
+
+describe('getRoomEventHandler', () => {
+  it('works', async () => {
```
opohjola, line 9: I would spec how it works

electron/channels/save-home-channel.spec.js
```diff
@@ -0,0 +1,24 @@
...
+HomeStore.mockImplementation(() => ({ set: mockHomeStoreSet }))
+
+describe('saveHomeEventHandler', () => {
+  it('calls store.set(home: params)', async () => {
+    const params = 'this is new home object to be saved'
+
+    const response = await eventHandlerWrapper('SAVE_HOME', saveHomeEventHandler)(mockEvent, params)
+
+    expect(mockHomeStoreSet.mock.calls.length).toBe(1)
+    expect(mockHomeStoreSet.mock.calls[0][0]).toEqual({ home: params })
```
opohjola, line 20: I typically use expect(mockHomeStoreSet).toHaveBeenCalledWith({ home: 'some-home' })

electron/channels/save-home-channel.spec.js
```diff
@@ -0,0 +1,24 @@
...
+// Mock electron-store
+jest.mock('electron-store')
+const mockHomeStoreSet = jest.fn()
+HomeStore.mockImplementation(() => ({ set: mockHomeStoreSet }))
+
+describe('saveHomeEventHandler', () => {
+  it('calls store.set(home: params)', async () => {
+    const params = 'this is new home object to be saved'
+
+    const response = await eventHandlerWrapper('SAVE_HOME', saveHomeEventHandler)(mockEvent, params)
```
opohjola, line 17: It is safer to not to use same variable in action and in expect; might pass even if it should not in some cases.

electron/channels/save-room-channel.js
```diff
@@ -1,17 +1,13 @@
...
-      console.log(JSON.stringify(diograph, null, 2))
-      event.reply(channels.SAVE_ROOM, true)
-    })
-    .catch((err) => {
-      console.log(err)
-      return event.reply(channels.SAVE_ROOM, null, err)
-    })
-})
+exports.saveRoomEventHandler = (event, { path, room: { id, diograph } }) =>
+  new Promise((resolve, reject) =>
```
opohjola, line 4: await saveRoom(path, diograph)
return true

electron/lib/room-util.js
```diff
@@ -1,9 +1,11 @@
 const fs = require('fs')
 const util = require('util')
 
+// No need to promisify: https://stackoverflow.com/a/52094177/1957884
+// - "Starting in Node v10 you can use await fs"
 const writeFilePromise = util.promisify(fs.writeFile)
 
-const getRoom = async (path) => {
+exports.getRoom = (path) => {
```
opohjola, line 8: I prefer exports in the end of file. Easier to see what is exported.

src/react/client/client.js
```diff
@@ -1,26 +1,48 @@
...
+ * @param Channel name
+ * @param Parameters
+ *
+ * @return Promise -> resolves with responseObject, rejects on Error object
+ *
+ * @example invokeChannel(channels.GENERATE_DIOGRAPH, address)
+ *
+ */
+export const invokeChannel = (channel, params) => {
+  // Development uses mock responses
```
opohjola, line 28: Why development uses mock responses?

src/react/client/client.spec.js
```diff
@@ -0,0 +1,52 @@
+import { ipcRenderer } from 'electron'
+import { invokeChannel } from './client'
+import { invokeAlertDialog } from './alertDialog'
+
+jest.mock('electron', () => ({ ipcRenderer: { invoke: jest.fn() } }), { virtual: true })
+jest.mock('./alertDialog', () => ({ invokeAlertDialog: jest.fn() }))
+
+const logInfoMock = jest.fn()
+const logErrorMock = jest.fn()
+window.frontendLogger = { info: logInfoMock, error: logErrorMock }
```
opohjola, line 10: This direct mocking might cause problems elsewhere. You might need to restore original behaviour or use some jest mocking / spy.

src/react/features/connectors/effects/useGetRoomEffect.js
```diff
@@ -4,20 +4,22 @@ import { updateConnection } from '../actions'
...
 
-export const useGetRoomEffect = (getRoomClient, connectorId) => {
+export const useGetRoomEffect = (connectorId) => {
   const { connect, disconnect } = useConnections(connectorId)
 
   const { dispatch, dispatchPromiseAction } = useDispatchActions()
   useEffect(() => {
     connect.forEach(({ address }) => {
       dispatchPromiseAction(
-        () => getRoomClient({ address }),
```
opohjola, line 15: This was a generic connector effect independent of folder connector. I would still separate this from folder connector specific methods.

src/react/features/connectors/effects/useSaveRoomEffect.js
```diff
@@ -4,16 +4,22 @@ import { useStore, useDispatchActions } from '../../../store'
...
+export const useSaveRoomEffect = (connectorId) => {
   const [{ diograph, updated }] = useStore((state) => state.diograph)
   const { connected } = useConnections(connectorId)
 
   const { debounceDispatchPromiseAction } = useDispatchActions()
   useEffect(() => {
     if (updated) {
       connected.forEach(({ address }) => {
-        debounceDispatchPromiseAction(() => saveRoomClient(address, { diograph }), saveRoom)
+        debounceDispatchPromiseAction(
```
opohjola, line 18: same comment as in getRoom

electron/lib/room-util.js
```diff
@@ -1,9 +1,11 @@
 const fs = require('fs')
 const util = require('util')
 
+// No need to promisify: https://stackoverflow.com/a/52094177/1957884
+// - "Starting in Node v10 you can use await fs"
```
opohjola, line 5: Why not to use await fs... instead of a comment?

opohjola (approved):
We need to align testing style. My testing opinions relate to almost all tests. However, no need to do it here but gradually.

The only change request is that we keep generic connectors independent of the folder connector specific methods.

electron/channels/channel-util.js
jvalanen: You mea  that `exports.` or the whole eventWrapper thing?

jvalanen: If this was different coding style, how would you have implemented the same functionality with your own style? E.g. having common logging and error handling

electron/channels/generate-diograph-channel.js
jvalanen: That's true. Although I didn't consider / touch that part here, I just wrapped it in promise and left the original implementation as is.

electron/channels/generate-diograph-channel.spec.js
jvalanen: I'm open for it just show me an example and I try to imitate. Didn't know that there's such templates or that the syntax matters.

electron/channels/get-room-channel.spec.js
jvalanen: Yeah, usually I start with 'works' just to get the first test and run the code. Sometimes that's already enough for now and when there's more complex idea behind the test that just "run the code". That also indicates that it is really just to run the code and nothing else, so the code coverage should be extended if you're about to do some bigger changes.

jvalanen: And as these channels didn't have any tests at all, this kind of quick "works" is 10x better than nothing at all.

electron/channels/save-home-channel.spec.js
jvalanen: Which variable you mean?

electron/channels/save-room-channel.js
jvalanen: The whole point is to return (a generic) Promise so the generic eventHandlerWrapper can solve it and deal with the success and reject its own way.

Although I'm not sure how async/await deals with rejections and if that could be applied and made this somehow better... 🤔

electron/lib/room-util.js
jvalanen: It was out if scope and wasn't sure about test coverage but wanted to add a FIXME note (without fixme prefix 😹) to remember that.

src/react/client/client.js
jvalanen: Because localhost:3300  it doesn't have backend...

jvalanen (commented):
Didn't have more time now but will send these to you already now if you happen to have time to respond some of the questions

src/react/features/connectors/effects/useGetRoomEffect.js
jvalanen: I don't understand this connector thing well enough to appreciate having this extra layer and those methods in [src/react/features/connectors/folder/client.js](https://github.com/DioryMe/diory-browser-electron/pull/79/files#diff-f36ecb3452677929abdba1db08152a2fc02201e9cd209e8e42d4f45241433a90)

To be honest I was (and still am :D ) pretty proud of myself that I was able to get my head around this thing enough to be able to make it so much simpler!

src/setupTests.js
```diff
@@ -0,0 +1,3 @@
+afterEach(() => {
+  jest.clearAllMocks()
```
jvalanen, line 2: Did you see this? 

I don't understand enough how (and why!) the mocks behave when defined a) before describe blocks or b) in `beforeEach` blocks but if they are cleared after each test, it should at least make the situation better... 🤔 

This way all the mocks need to be re-defined in each test and tests shouldn't distract each other... (at least according to my best understanding...)

src/react/client/client.spec.js
jvalanen: The restoring is done in `afterEach` block in [setupTests.js](https://github.com/DioryMe/diory-browser-electron/pull/79/files#diff-f1995b7576ce098f06844c16df12561f4694c56c84869ec5180933e652f56862)

jvalanen: It is done in global `afterEach` block, see: https://github.com/DioryMe/diory-browser-electron/pull/79/files#diff-f1995b7576ce098f06844c16df12561f4694c56c84869ec5180933e652f56862

Although I'm not sure what it is exactly doing but that's the solution I ended up to...

PR #84, opened 2021-01-17
DDA-80: Verify room connections

jvalanen:
--- WIP because target branch `DDA-000-alert-to-frontend` hasn't been merged yet. ---

1. Adds `errors` key to GET_HOME channel
2. Verifies that each connectionURI file path exists if not adds that connectionURI to `errors` in response
3. Frontend invokes alert for each string in GET_HOME response's `errors`


electron/channels/get-home-channel.js
```diff
@@ -31,11 +32,27 @@ exports.getHomeEventHandler = (event, params) => {
...
+
   return new Promise((resolve, reject) => {
     resolve(home)
   })
 }
 
+/*
+ * Returns true if roomConnection is ok and false if not
+ */
+exports.roomConnectionOK = (connectionURI, connectionConfig) => fs.existsSync(connectionURI)
```
opohjola, line 54: we could use "roomConnectionIsOk" -syntax for methods that return boolean.

electron/channels/get-home-channel.js
```diff
@@ -31,11 +32,27 @@ exports.getHomeEventHandler = (event, params) => {
   const store = new HomeStore()
   const home = store.get('home') || defaultHome
 
+  home.errors = []
+  Object.keys(home.connections).forEach((connectionURI) => {
+    const roomConnectionOK = exports.roomConnectionOK(
+      connectionURI,
```
opohjola, line 38: use always same case = camelCase

electron/channels/get-home-channel.js
```diff
@@ -31,11 +32,27 @@ exports.getHomeEventHandler = (event, params) => {
   const store = new HomeStore()
   const home = store.get('home') || defaultHome
 
+  home.errors = []
+  Object.keys(home.connections).forEach((connectionURI) => {
+    const roomConnectionOK = exports.roomConnectionOK(
```
opohjola, line 37: no need for exports if roomConnectionOK is before this method (or it has traditional function syntax)

electron/channels/get-home-channel.js
```diff
@@ -31,11 +32,27 @@ exports.getHomeEventHandler = (event, params) => {
...
+      home.errors.push({ connectionURI })
+    }
+  })
+
   return new Promise((resolve, reject) => {
     resolve(home)
   })
 }
 
+/*
```
opohjola, line 51: roomConnectionIsOk says the same; no need for comment

electron/channels/get-home-channel.spec.js
```diff
@@ -1,14 +1,40 @@
 const HomeStore = require('electron-store')
 const { eventHandlerWrapper } = require('./channel-util')
-const { getHomeEventHandler, defaultHome } = require('./get-home-channel')
+const getHomeChannel = require('./get-home-channel')
+
+const { getHomeEventHandler, defaultHome } = getHomeChannel
```
opohjola, line 5: Why not in require?

electron/channels/get-home-channel.spec.js
```diff
@@ -1,14 +1,40 @@
...
+    roomId: 'development-content-room',
+    dioryId: 'development-content-room',
+  },
+  errors: [],
+}
+
 describe('getHomeEventHandler', () => {
+  beforeEach(() => {
+    // Mock roomConnectionOK
+    getHomeChannel.roomConnectionOK = jest.fn(() => true)
```
opohjola, line 35: there is some code smell here when there is a need to mock internal function.

electron/channels/get-home-channel.js
opohjola: meaning 'is' in name

electron/channels/get-home-channel.spec.js
```diff
@@ -19,11 +45,25 @@ describe('getHomeEventHandler', () => {
...
+    // Mock roomConnectionOK to return false
+    getHomeChannel.roomConnectionOK = jest.fn(() => false)
 
     const params = {}
     const response = await eventHandlerWrapper('GET_HOME', getHomeEventHandler)(mockEvent, params)
 
-    expect(response).toEqual('value')
+    // Expect 'errors' to be added to the response
+    const homeJSONWithErrors = homeJSON
+    homeJSONWithErrors.errors = ['./public/development-content-room']
```
opohjola, line 66: this changes also homeJSON and might cause problems in other tests.
It is safer to set data in beforeEach, i.e., always begin a test with empty data.

src/react/features/rooms/effects/useGetHomeEffect.js
```diff
@@ -7,12 +7,22 @@ import { setConnections } from '../../connectors/actions'
...
   const dispatch = useDispatch()
 
   useEffect(() => {
-    invokeChannel(channels.GET_HOME).then(({ rooms, connections, focus }) => {
+    invokeChannel(channels.GET_HOME).then(({ rooms, connections, focus, errors }) => {
+      errors.forEach((error) => {
+        invokeAlertDialog(
+          `Couldn't connect to room ${error.connectionURI}, path didn't exist. Please remove and re-add the room to reconnect.`
+        )
+      })
```
opohjola, line 25: We would add error feature with actions, reducer and components. 
Then here dispatch(addErrors(errors)))

src/react/features/rooms/effects/useGetHomeEffect.js
```diff
@@ -7,12 +7,22 @@ import { setConnections } from '../../connectors/actions'
 
 import { invokeChannel } from '../../../client/client'
 import { channels } from '../../../../shared/constants'
-
+import { invokeAlertDialog } from '../../../client/alertDialog'
+/*
+ * Requests home & focus info via GET_HOME and sets the focus for room & diory
```
opohjola, line 12: I'm not in favour for comments. The code itself or specs (tests) at least should say all this.

opohjola: If they does not, they should be improved. The need for comment typically means that the code is not good enough.

src/react/features/rooms/effects/useGetHomeEffect.js
```diff
@@ -7,12 +7,22 @@ import { setConnections } from '../../connectors/actions'
...
+ * - saves the response to the store
+ * - invoke an alert for each error
+ */
 export const useGetHomeEffect = () => {
   const dispatch = useDispatch()
 
   useEffect(() => {
-    invokeChannel(channels.GET_HOME).then(({ rooms, connections, focus }) => {
+    invokeChannel(channels.GET_HOME).then(({ rooms, connections, focus, errors }) => {
```
opohjola, line 20: invokeChannel(channels.GET_HOME) is something that might require a specific method with a good name for clarity here. getHomeDataFromBackendStore?

src/react/features/rooms/effects/useGetHomeEffect.js
opohjola: Comments will typically be obsolete.

src/react/features/rooms/effects/useGetHomeEffect.js
opohjola: should we have a method for all invokeChannel(channels...)s?

electron/channels/get-home-channel.spec.js
```diff
@@ -19,11 +45,25 @@ describe('getHomeEventHandler', () => {
...
+
+    const params = {}
+    const response = await eventHandlerWrapper('GET_HOME', getHomeEventHandler)(mockEvent, params)
+
+    expect(response).toEqual(homeJSON)
+  })
+
+  it('invalid file path', async () => {
+    HomeStore.mockImplementation(() => ({ get: () => homeJSON }))
+    // Mock roomConnectionOK to return false
```
opohjola, line 58: The test code should say this. No need for comment.

opohjola: describe: given room path is not ok,
it: returns home with room path in errors

opohjola (approved):
Mostly and testing coding style comments; some are in imperative but are just my opinions.

Some test setups might cause problems later but likely not now.

electron/channels/get-home-channel.spec.js
jvalanen: I think somehow the mocks didn't work without splitting this on two lines...

getHomeChannel is needed to be able to mock roomConnectionOK

opohjola: it's not good practise to mock internal functions; mock  fs.existsSync(connectionURI) instead.

electron/channels/get-home-channel.js
jvalanen: But that's an acronym? (universal resource identifier)
And I thought OK is acronym too... (although ok or Ok are fine too...)

So I would say that's correct camelCase

electron/channels/get-home-channel.spec.js
jvalanen: Yeah, that's actually true. But I may have thought that this function might get more complicated and I don't want to break my test by depending on its implementation.

And here again it's a matter of defining interfaces:
- a) if roomConnectionOK is a private function we shouldn't mock anything but make a test that actually hits the file system
- b) if roomConnectionOK is a public interface, this is exactly what we want (although we should write a test for roomConnectionOK itself elsewhere)

electron/channels/get-home-channel.spec.js
jvalanen: It says room_Connection_ . Currently it is checking the path but the big idea behind it that "checking the room connection" might be much more complex operation and depend on the connection type etc. (if you are offline and room requires wifi to connect, then this should return false)

But yes, you are right it is an unnecessary comment here, should be deleted.

src/react/features/rooms/effects/useGetHomeEffect.js
jvalanen: I agree! But if you know the code is not currently clear enough and don't know how to or don't have time to fix it now, it's better to leave a comment which explains it better (than just leave it without any clarification, better to over-communicate!!)

electron/channels/get-home-channel.spec.js
jvalanen: Okay...that's interesting.
If I wanted to do this should I do something like 
```
const homeJSONWithErrors = JSON.parse(JSON.stringify(homeJSON))
```

🤔

src/react/features/rooms/effects/useGetHomeEffect.js
jvalanen: That's something we can think about.
Once we have good test coverage and figured out good names , adding that kind of layer should be a no-brainer.

jvalanen (commented):
Thanks for the comments! 
Let's continue these PR discussions and little by little we understand each other better and better.  And that way we will reach shared conventions and coding style.

I won't make any changes for now as the overall understanding and common vision is still a bit blurry and we need to make big changes to these in the future anyway. Let's fix them then at once!

PR #93, opened 2021-02-05
DDA-85: E2E tests + read diograph.json as default

jvalanen:
**Biggest changes:**
- E2E tests with Testcafe 
    - uusi diograph folder
    - diograph folder
    - snapshottien ottaminen
- Load diograph folder by reading diograph.json if found
- New diograph format with `rootId`

**Smaller changes:**
- Rename get/saveRoom to read/saveDiographJSON
- Remove `util.promisify`
- Testi fixturet konvention mukaan
- describe-blokkeja lisätty

### DIFF
--- a/electron/channels/generate-diograph-channel.js
+++ b/electron/channels/generate-diograph-channel.js
@@ -26,7 +26,7 @@ exports.generateDiographEventHandler = (event, path) =>
       resolve({ rootId, diograph, path })
     }
     generateDiograph(path).then(({ rootId, diograph }) => {
-      saveDiographJSON(path, diograph)
+      saveDiographJSON(path, diograph, rootId)
         .then(() => {
           resolve({ rootId, diograph, path })
         })
--- a/electron/channels/generate-diograph-channel.js
+++ b/electron/channels/generate-diograph-channel.js
@@ -10,7 +10,7 @@ const { saveDiographJSON, readDiographJSON } = require('../lib/room-util')
  *
  * @example Response object:
  * {
- *   id: 'diory2',
+ *   rootId: 'diory2',
  *   path: '/Users/...',
  *   diograph: {
  *     diory1: { id: 'diory1', image: '...', links: [...] },
--- a/electron/channels/generate-diograph-channel.js
+++ b/electron/channels/generate-diograph-channel.js
@@ -13,8 +13,8 @@ const { saveRoom } = require('../lib/room-util')
  *   id: 'diograph123',
  *   path: '/Users/...',
  *   diograph: {
- *      room1: { id: 'room1', image: '...', links: [...] },
- *      room2: { id: 'room2', image: '...', links: [...] }
+ *     diory1: { id: 'diory1', image: '...', links: [...] },
+ *     diory2: { id: 'diory2', image: '...', links: [...] }
  *   }
  * }
  *
--- a/electron/channels/generate-diograph-channel.js
+++ b/electron/channels/generate-diograph-channel.js
@@ -24,7 +24,7 @@ exports.generateDiographEventHandler = (event, path) =>
     const { diograph } = readDiographJSON(path)
     if (diograph) {
       resolve({
-        id: 'some-id',
+        id: Object.keys(diograph)[0],
         diograph,
         path,
       })
--- a/electron/channels/generate-diograph-channel.js
+++ b/electron/channels/generate-diograph-channel.js
@@ -21,18 +21,14 @@ const { saveDiographJSON, readDiographJSON } = require('../lib/room-util')
  */
 exports.generateDiographEventHandler = (event, path) =>
   new Promise((resolve, reject) => {
-    const { diograph } = readDiographJSON(path)
+    const { rootId, diograph } = readDiographJSON(path)
     if (diograph) {
-      resolve({
-        id: Object.keys(diograph)[0],
-        diograph,
-        path,
-      })
+      resolve({ rootId, diograph, path })
     }
-    generateDiograph(path).then(({ id, diograph }) => {
+    generateDiograph(path).then(({ rootId, diograph }) => {
       saveDiographJSON(path, diograph)
         .then(() => {
-          resolve({ id, diograph, path })
+          resolve({ rootId, diograph, path })
         })
         .catch((err) => {
           reject(err)
--- a/electron/channels/generate-diograph-channel.js
+++ b/electron/channels/generate-diograph-channel.js
@@ -24,14 +24,15 @@ exports.generateDiographEventHandler = (event, path) =>
     const { rootId, diograph } = readDiographJSON(path)
     if (diograph) {
       resolve({ rootId, diograph, path })
+    } else {
+      generateDiograph(path).then(({ rootId, diograph }) => {
+        saveDiographJSON(path, diograph, rootId)
+          .then(() => {
+            resolve({ rootId, diograph, path })
+          })
+          .catch((err) => {
+            reject(err)
+          })
+      })
     }
-    generateDiograph(path).then(({ rootId, diograph }) => {
-      saveDiographJSON(path, diograph, rootId)
-        .then(() => {
-          resolve({ rootId, diograph, path })
-        })
-        .catch((err) => {
-          reject(err)
-        })
-    })
   })
--- a/electron/channels/generate-diograph-channel.js
+++ b/electron/channels/generate-diograph-channel.js
@@ -1,5 +1,5 @@
 const { generateDiograph } = require('../generators/diograph-generator')
-const { saveDiographJSON, readDiographJSON } = require('../lib/room-util')
+const { saveDiographJson, readDiographJson } = require('../lib/room-util')
 
 /**
  * Event handler for GENERATE_DIOGRAPH channel
@@ -21,12 +21,12 @@ const { saveDiographJSON, readDiographJSON } = require('../lib/room-util')
  */
 exports.generateDiographEventHandler = (event, path) =>
   new Promise((resolve, reject) => {
-    const { rootId, diograph } = readDiographJSON(path)
+    const { rootId, diograph } = readDiographJson(path)
     if (diograph) {
       resolve({ rootId, diograph, path })
     } else {
       generateDiograph(path).then(({ rootId, diograph }) => {
-        saveDiographJSON(path, diograph, rootId)
+        saveDiographJson(path, diograph, rootId)
           .then(() => {
             resolve({ rootId, diograph, path })
           })
--- a/electron/channels/generate-diograph-channel.js
+++ b/electron/channels/generate-diograph-channel.js
@@ -1,5 +1,6 @@
 const { generateDiograph } = require('../generators/diograph-generator')
-const { saveDiographJson, readDiographJson } = require('../lib/room-util')
+const { saveDiographJson } = require('../lib/save-diograph-json')
+const { readDiographJson } = require('../lib/read-diograph-json')
 
 /**
  * Event handler for GENERATE_DIOGRAPH channel
--- a/electron/channels/generate-diograph-channel.js
+++ b/electron/channels/generate-diograph-channel.js
@@ -1,5 +1,5 @@
 const { generateDiograph } = require('../generators/diograph-generator')
-const { saveRoom } = require('../lib/room-util')
+const { saveDiographJSON } = require('../lib/room-util')
 
 /**
  * Event handler for GENERATE_DIOGRAPH channel
@@ -10,7 +10,7 @@ const { saveRoom } = require('../lib/room-util')
  *
  * @example Response object:
  * {
- *   id: 'diograph123',
+ *   id: 'diory2',
  *   path: '/Users/...',
  *   diograph: {
  *     diory1: { id: 'diory1', image: '...', links: [...] },
@@ -22,7 +22,7 @@ const { saveRoom } = require('../lib/room-util')
 exports.generateDiographEventHandler = (event, path) =>
   new Promise((resolve, reject) => {
     generateDiograph(path).then(({ id, diograph }) => {
-      saveRoom(path, diograph)
+      saveDiographJSON(path, diograph)
         .then(() => {
           resolve({ id, diograph, path })
         })
--- a/electron/channels/generate-diograph-channel.js
+++ b/electron/channels/generate-diograph-channel.js
@@ -21,11 +21,11 @@ const { saveDiographJSON, readDiographJSON } = require('../lib/room-util')
  */
 exports.generateDiographEventHandler = (event, path) =>
   new Promise((resolve, reject) => {
-    const diographJSON = readDiographJSON(path)
-    if (diographJSON) {
+    const { diograph } = readDiographJSON(path)
+    if (diograph) {
       resolve({
-        id: 'id',
-        diograph: diographJSON,
+        id: 'some-id',
+        diograph,
         path,
       })
     }
--- a/electron/channels/generate-diograph-channel.js
+++ b/electron/channels/generate-diograph-channel.js
@@ -1,5 +1,5 @@
 const { generateDiograph } = require('../generators/diograph-generator')
-const { saveDiographJSON } = require('../lib/room-util')
+const { saveDiographJSON, readDiographJSON } = require('../lib/room-util')
 
 /**
  * Event handler for GENERATE_DIOGRAPH channel
@@ -21,13 +21,23 @@ const { saveDiographJSON } = require('../lib/room-util')
  */
 exports.generateDiographEventHandler = (event, path) =>
   new Promise((resolve, reject) => {
-    generateDiograph(path).then(({ id, diograph }) => {
-      saveDiographJSON(path, diograph)
-        .then(() => {
-          resolve({ id, diograph, path })
+    // WHY readDiographJSON returns a promise?!!?
+    readDiographJSON(path).then((diographJSON) => {
+      if (diographJSON) {
+        resolve({
+          id: 'id',
+          diograph: diographJSON,
+          path,
         })
-        .catch((err) => {
-          reject(err)
-        })
+      }
+      generateDiograph(path).then(({ id, diograph }) => {
+        saveDiographJSON(path, diograph)
+          .then(() => {
+            resolve({ id, diograph, path })
+          })
+          .catch((err) => {
+            reject(err)
+          })
+      })
     })
   })
--- a/electron/channels/generate-diograph-channel.js
+++ b/electron/channels/generate-diograph-channel.js
@@ -21,23 +21,21 @@ const { saveDiographJSON, readDiographJSON } = require('../lib/room-util')
  */
 exports.generateDiographEventHandler = (event, path) =>
   new Promise((resolve, reject) => {
-    // WHY readDiographJSON returns a promise?!!?
-    readDiographJSON(path).then((diographJSON) => {
-      if (diographJSON) {
-        resolve({
-          id: 'id',
-          diograph: diographJSON,
-          path,
-        })
-      }
-      generateDiograph(path).then(({ id, diograph }) => {
-        saveDiographJSON(path, diograph)
-          .then(() => {
-            resolve({ id, diograph, path })
-          })
-          .catch((err) => {
-            reject(err)
-          })
+    const diographJSON = readDiographJSON(path)
+    if (diographJSON) {
+      resolve({
+        id: 'id',
+        diograph: diographJSON,
+        path,
       })
+    }
+    generateDiograph(path).then(({ id, diograph }) => {
+      saveDiographJSON(path, diograph)
+        .then(() => {
+          resolve({ id, diograph, path })
+        })
+        .catch((err) => {
+          reject(err)
+        })
     })
   })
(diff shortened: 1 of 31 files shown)


electron/channels/generate-diograph-channel.js
```diff
@@ -1,5 +1,5 @@
 const { generateDiograph } = require('../generators/diograph-generator')
-const { saveRoom } = require('../lib/room-util')
+const { saveDiographJSON, readDiographJSON } = require('../lib/room-util')
```
opohjola: I propose that we use camelCase in variables and function names. Mixed case causes problems with replace.

opohjola: Rename room-util. I don't like this something -convention but it is used commonly in node.js. Could we just have files with same name as the function: saveDiograph.js and getDiograph.js would be my proposals here.

electron/channels/generate-diograph-channel.js
```diff
@@ -10,24 +10,29 @@ const { saveRoom } = require('../lib/room-util')
...
- *      room1: { id: 'room1', image: '...', links: [...] },
- *      room2: { id: 'room2', image: '...', links: [...] }
+ *     diory1: { id: 'diory1', image: '...', links: [...] },
+ *     diory2: { id: 'diory2', image: '...', links: [...] }
  *   }
  * }
  *
  */
 exports.generateDiographEventHandler = (event, path) =>
   new Promise((resolve, reject) => {
```
opohjola, line 24: Should we use async - await?

electron/channels/generate-diograph-channel.js
```diff
@@ -10,24 +10,29 @@ const { saveRoom } = require('../lib/room-util')
...
  *   diograph: {
- *      room1: { id: 'room1', image: '...', links: [...] },
- *      room2: { id: 'room2', image: '...', links: [...] }
+ *     diory1: { id: 'diory1', image: '...', links: [...] },
+ *     diory2: { id: 'diory2', image: '...', links: [...] }
  *   }
  * }
  *
  */
 exports.generateDiographEventHandler = (event, path) =>
```
opohjola, line 23: I propose named function and exports in the end of file.

electron/channels/generate-diograph-channel.js
```diff
@@ -10,24 +10,29 @@ const { saveRoom } = require('../lib/room-util')
...
-        })
-        .catch((err) => {
-          reject(err)
-        })
-    })
+    const { rootId, diograph } = readDiographJSON(path)
+    if (diograph) {
+      resolve({ rootId, diograph, path })
+    } else {
+      generateDiograph(path).then(({ rootId, diograph }) => {
```
opohjola, line 29: with await this would be:
const { rootId, diograph } = readDiographJSON(path)
if (diograph) {
  return { rootId, diograph, path }
}
const { rootId, diograph } = await generateDiograph(path)
await saveDiographJSON(path, diograph, rootId)
return { rootId, diograph, path }

opohjola: or even better:
let { rootId, diograph } = readDiographJSON(path)
if (!diograph) {
 let { rootId, diograph } = await generateDiograph(path)
 await saveDiographJSON(path, diograph, rootId)
}
return { rootId, diograph, path }

electron/channels/generate-diograph-channel.spec.js
```diff
@@ -1,44 +1,77 @@
...
 const { generateDiograph } = require('../generators/diograph-generator')
-const { saveRoom } = require('../lib/room-util')
+const { saveDiographJSON, readDiographJSON } = require('../lib/room-util')
 
 // Mock event.reply
 const mockEventReply = jest.fn()
 const mockEvent = { reply: mockEventReply }
 // Mock generateDiograph
 jest.mock('../generators/diograph-generator')
 const generateDiographMock = generateDiograph.mockResolvedValue({
```
opohjola: I advise to set mocks in beforeEach

electron/channels/generate-diograph-channel.spec.js
```diff
@@ -1,44 +1,77 @@
...
-  id: 'this',
-  diograph: 'is generated diograph',
+  rootId: 'some-diory-id',
```
opohjola: I propose to use when possible attribute name in variable: rootId: 'some-rootId'. This makes renaming easier and causes less errors.

electron/channels/generate-diograph-channel.spec.js
```diff
@@ -1,44 +1,77 @@
...
-    expect(generateDiograph).toHaveBeenCalledWith(params)
-
-    expect(saveRoom).toHaveBeenCalledTimes(1)
-    expect(saveRoom).toHaveBeenCalledWith(params, 'is generated diograph')
-
-    const responseObject = {
-      id: 'this',
-      diograph: 'is generated diograph',
-      path: 'this is a file path',
+  describe('if diograph.json not found', () => {
```
opohjola, line 16: My proposal for specs would be:
given path to a folder
it reads diograph from folder
when diograph is found
it resolves diograph, diograph root id and path
when diograph is not found
it generates diograph from the folder
when diograph generation succeed
it saves diograph
it resolves diograph, diograph root id and path
when diograph generation fails
it rejects with error

opohjola: this is the style that I use in all my projects.

opohjola: This set all the mocks and values in respective step.
Typically there is a pattern of 
describe
__beforeEach
__it
__it
__describe
____beforeEach
____it
____it

electron/channels/get-room-channel.spec.js
```diff
@@ -14,7 +14,7 @@ describe('getRoomEventHandler', () => {
 
     const diographJSONRawContents = fs.readFileSync(`${exampleFolderPath}/diograph.json`)
     const diographObject = JSON.parse(diographJSONRawContents)
```
opohjola: I propopse to organize test rows with  AAA:
Assign
(empty line)
Act
(empty line)
Assert

electron/channels/save-room-channel.spec.js
```diff
@@ -1,43 +1,65 @@
...
 
 // Mock event.reply
 const mockEventReply = jest.fn()
 const mockEvent = { reply: mockEventReply }
-// Mock saveRoom
+// Mock saveDiographJSON
 jest.mock('../lib/room-util')
 
 const params = {
-  path: 'this is a file path',
```
opohjola, line 12: I advise to populate params in beforeEach

electron/channels/save-room-channel.js
```diff
@@ -1,13 +1,13 @@
-const { saveRoom } = require('../lib/room-util')
+const { saveDiographJSON } = require('../lib/room-util')
 
-exports.saveRoomEventHandler = (event, { path, room: { id, diograph } }) =>
-  new Promise((resolve, reject) =>
-    saveRoom(path, diograph).then(
+exports.saveRoomEventHandler = (event, { path, room: { rootId, diograph } }) =>
```
opohjola, line 3: Should we rename to saveDiographEventHandler?

electron/channels/save-room-channel.spec.js
```diff
@@ -1,43 +1,65 @@
...
-  it("'true, undefined' if Promise resolves", async () => {
-    const saveRoomMock = saveRoom.mockResolvedValue(undefined)
+describe('saveRoomEventHandler', () => {
+  describe('when saveDiographJSON resolves properly', () => {
+    let saveDiographJSONMock
+    beforeEach(() => {
+      saveDiographJSONMock = saveDiographJSON.mockResolvedValue(undefined)
+    })
+    it('returns true', async () => {
+      const response = await eventHandlerWrapper('SAVE_ROOM', saveRoomEventHandler)(
```
opohjola, line 29: Why eventHandlerWrapper is needed here?

electron/channels/save-room-channel.js
```diff
@@ -1,13 +1,13 @@
...
+  new Promise((resolve, reject) => {
```
opohjola, line 4: Exactly the same with await:
await saveDiographJSON(path, diograph, rootId)
return true

electron/channels/save-room-channel.js
opohjola: and remove room from props

electron/channels/save-room-channel.spec.js
```diff
@@ -1,43 +1,65 @@
...
+    rootId: 'root-diory-id',
+    diograph: 'some-diograph',
   },
 }
 
-describe('saveRoomEventHandler sends event with payload ', () => {
-  it("'true, undefined' if Promise resolves", async () => {
-    const saveRoomMock = saveRoom.mockResolvedValue(undefined)
+describe('saveRoomEventHandler', () => {
+  describe('when saveDiographJSON resolves properly', () => {
```
opohjola: my spec would be:
given diograph, diograph root id and path
it saves diograph to a folder on path
when saving succeed
it returns true
when saving fails
it returns saving error

electron/channels/save-room-channel.spec.js
```diff
@@ -1,43 +1,65 @@
...
+    let saveDiographJSONMock
```
opohjola: saveDiographJSONMock should not be needed...
saveDiographJSON.mockResolvedValue(undefined) should be enough

electron/channels/save-room-channel.spec.js
```diff
@@ -1,43 +1,65 @@
...
+    beforeEach(() => {
+      saveDiographJSONMock = saveDiographJSON.mockResolvedValue(undefined)
+    })
+    it('returns true', async () => {
+      const response = await eventHandlerWrapper('SAVE_ROOM', saveRoomEventHandler)(
+        mockEvent,
+        params
+      )
+      await saveDiographJSONMock
```
opohjola: should not be needed...

electron/channels/save-room-channel.spec.js
```diff
@@ -1,43 +1,65 @@
...
 
-  it("'null, err' if Promise rejects", async () => {
-    const saveRoomMock = saveRoom.mockRejectedValue(new Error('this is an error'))
-
-    const response = await eventHandlerWrapper('SAVE_ROOM', saveRoomEventHandler)(mockEvent, params)
-    await saveRoomMock
+  describe('when saveDiographJSON rejects', () => {
+    let saveDiographJSONMock
+    beforeEach(() => {
+      saveDiographJSONMock = saveDiographJSON.mockRejectedValue(new Error('some-error'))
```
opohjola: saveDiographJSON.mockRejectedValue(new Error('some-error')) should be enough

electron/lib/room-util.js
```diff
@@ -1,32 +1,59 @@
...
     const raw = fs.readFileSync(diographJSONPath)
-    return {
-      diograph: JSON.parse(raw),
-    }
+    const { rootId, diograph } = JSON.parse(raw)
+    return { rootId, diograph }
+  }
+
+  backendLogger.info('readDiographJSON: diograph.json not found in', path)
+  return {
```
opohjola: Early return in error cases could be a good convention.

opohjola: for some reason returning undefined do not feel right but I couldn't find a reason for the feeling. null, empty object?

src/react/features/connectors/effects/useGetRoomEffect.js
```diff
@@ -12,10 +12,10 @@ export const useGetRoomEffect = (connectorId) => {
 
   const { dispatch, dispatchPromiseAction } = useDispatchActions()
   useEffect(() => {
-    connect.forEach(({ address }) => {
+    connect.forEach(({ rootId, diograph, address }) => {
```
opohjola, line 15: connect should not have diograph... maybe not event rootId?? 

invokeChannel(channels.GET_ROOM, { address }) returns { rootId, diograph } right?

opohjola: dispatchPromiseAction should be improved; it's not good enough now.

src/react/features/connectors/effects/useGetRoomEffect.js
```diff
@@ -12,10 +12,10 @@ export const useGetRoomEffect = (connectorId) => {
```
opohjola, line 12: should this be renamed to useGetDiographEffect?

src/react/features/connectors/effects/useGetRoomEffect.js
```diff
@@ -12,10 +12,10 @@ export const useGetRoomEffect = (connectorId) => {
 
   const { dispatch, dispatchPromiseAction } = useDispatchActions()
   useEffect(() => {
-    connect.forEach(({ address }) => {
+    connect.forEach(({ rootId, diograph, address }) => {
       dispatchPromiseAction(
         () => invokeChannel(channels.GET_ROOM, { address }),
```
opohjola, line 17: I propose that all channels are moved to client that have methods like invokeGetRoomChannel of something...

src/react/features/connectors/effects/useSaveRoomEffect.js
```diff
@@ -8,15 +8,15 @@ import { invokeChannel } from '../../../client/client'
...
+  const [{ rootId, diograph, updated }] = useStore((state) => state.diograph)
   const { connected } = useConnections(connectorId)
 
   const { debounceDispatchPromiseAction } = useDispatchActions()
   useEffect(() => {
     if (updated) {
       connected.forEach(({ address }) => {
         debounceDispatchPromiseAction(
-          () => invokeChannel(channels.SAVE_ROOM, { path: address, room: { diograph } }),
+          () => invokeChannel(channels.SAVE_ROOM, { path: address, room: { rootId, diograph } }),
```
opohjola, line 19: This is correct!

src/react/features/connectors/folder/buttons/useAddConnectionButton.js
```diff
@@ -20,19 +20,19 @@ export const useAddConnectionButton = () => {
   useEffect(() => {
     const getHome = (result) => {
       const address = result.filePaths[0]
-      invokeChannel(channels.GENERATE_DIOGRAPH, address).then(({ id, diograph }) => {
+      invokeChannel(channels.GENERATE_DIOGRAPH, address).then(({ rootId, diograph }) => {
```
opohjola, line 23: Should we actually return just root diory from GENERATE_DIOGRAPH? That is the only thing that is used now.

src/react/features/connectors/folder/buttons/useAddConnectionButton.js
```diff
@@ -20,19 +20,19 @@ export const useAddConnectionButton = () => {
...
         const roomId = uuid()
         dispatch(addConnection({ address, room: roomId, connector: 'file' }))
-        dispatch(addRoom(roomId, diograph[id]))
+        dispatch(addRoom(roomId, diograph[rootId]))
         dispatch(enterRoom({ id: roomId }))
-        dispatch(setFocus({ focus: id }))
+        dispatch(setFocus({ focus: rootId }))
```
opohjola, line 28: As root id comes now from getRoom (-> getDiograph), I thing set focus should happen there.

src/react/features/connectors/effects/useGetRoomEffect.js
```diff
@@ -12,10 +12,10 @@ export const useGetRoomEffect = (connectorId) => {
...
-        () => getRoom({ address })
+        () => getRoom({ rootId, diograph, address })
       )
```
opohjola, line 19: We could set focus here (I hope):
.then(({ rootId }) => dispatch(setFocus({ focus: rootId })

src/react/features/connectors/folder/buttons/useAddConnectionButton.js
opohjola: in useGetRoomEffect

opohjola (commented):
Lot of proposals, useGetRoomEffect should be fixed, await would save a lot of lines (a good thing)

src/react/features/connectors/effects/useGetRoomEffect.js
jvalanen: I don't see particular need for change in here. I think those invokeChannel stuff are clear enough.

But if we want to change it, I'd propose: `channelClient.getRoom({ address })` (and similar for all the other channels)

src/react/features/connectors/effects/useGetRoomEffect.js
jvalanen: I don't have good enough overall view about this architecture to say anything about this :/
Maybe we could or maybe current is good enough? 🙄 

My current goal is just to 
1) make progress in feature-wise
2) improve the test coverage so if we want to change something, refactoring would be nice and easy against green tests that won't break (concretely the E2E tests)
3) Little by little get a good grasp of the current code base and be able to propose / have something to say about architectural stuff. Until then I take all the existing stuff as granted and "good" because I don't know anything better...

src/react/features/connectors/effects/useGetRoomEffect.js
jvalanen: Unfortunately I don't know the idea behind connectors well enough to say anything about this.

electron/channels/generate-diograph-channel.js
jvalanen: We need to form a convention here.

My things for this convetion is:
1) In frontend/React it is already like this: `export const useButtonBar = () => {`
2) For me it is important to see if this is "the main function of the file" or just a helper. If it is in the end of the file, it's very hard to find out.

jvalanen: If you mean this by "named functions", that's ok for me:
```
exports.generateDiographEventHandler = function (event, path) {
```

But still it conflicts with the React convention and I think we should have the same conventions in front and backend (as much as possible)

electron/channels/generate-diograph-channel.js
jvalanen: If you mean `saveDiographJson`, it's ok for me and I can change these.

But did you read about my reasons behind this form? And my follow-up-comments in general here: https://github.com/DioryMe/diory-browser-electron/pull/84#discussion_r569607958

jvalanen: I think it is weird too but it has been like that as long as I remember. So I thought you have some reason to have it that way 😄 

I can change those. Split them to their own files.

electron/channels/generate-diograph-channel.js
jvalanen: It's a same reason here than in my previous comments: https://github.com/DioryMe/diory-browser-electron/pull/79#discussion_r569173042 => currently these `xxxEventHandler`s are designed/built to return promises, that way it's easy to have common logging and error handling. (may be easy to make the same thing with await but didn't thought about it and wanted to change the code as little as possible when doing the channel refactoring...)

electron/channels/generate-diograph-channel.js
jvalanen: Hint: Use code-markup for better readability with this notation: ```  or "insert code" from toolbar.

```
then it looks like this
```

electron/channels/get-room-channel.spec.js
jvalanen: Isn't it already like that?

Assign
```
    const exampleFolderPath = './public/development-content-room'
    const params = { address: exampleFolderPath }
```

Act:
```
    const response = await eventHandlerWrapper('GET_ROOM', getRoomEventHandler)(mockEvent, params)
```

Assert:
```
    const diographJSONRawContents = fs.readFileSync(`${exampleFolderPath}/diograph.json`)
    const diographObject = JSON.parse(diographJSONRawContents)
    const returnValue = { rootId: diographObject.rootId, diograph: diographObject.diograph }
    --- (in here there's an excessive empty line, I admit :) ) ---
    expect(response).toEqual(returnValue)
```

electron/channels/save-room-channel.js
jvalanen: I propose having a bigger design/architecture debate about these channels: 
1) Having a `channelClient` in frontend
2) Using await in backend
3) Renaming stuff: save/getRoom -> save/getDiograph etc.
4) Fine-tuning the request arguments and response keys
5) Should we have frontend to call `saveRoom` explicitly after calling generateDiograph? (currently backend calls first generateDiograph and then saveRoom)

I would like to have a big picture in line with everything before making any changes (as it is working pretty well feature-wise now and doesn't need urgent changes)

electron/channels/save-room-channel.spec.js
jvalanen: Aah, okay, I'll try it

electron/channels/save-room-channel.spec.js
jvalanen: Because that's what the SAVE_ROOM channel calls when it is invoked from the frontend? 

All the channel-specs are done the same way.

electron/channels/generate-diograph-channel.js
opohjola: Yes, correct. It is acronyme but more important is to able rename fast and without errors. That requires same case. E.g. changing with replaceAll JSON to file: saveDiographJSON -> saveDiographFILE; saveDiographJson -> saveDiographFile

electron/channels/generate-diograph-channel.js
opohjola: async await returns promise!!!

jvalanen: Aah, oh! 🤔 😄  Didn't know that but makes sense! 😊 

But I'll leave this change to wait until we know exactly what we want to do to the channels (both in front & backend) and then it's easy to refactor it everywhere as we already have the test coverage in place. So we don't waste time changing a working code several times just for nothing...

PR #95, opened 2021-02-16
WIP: POC: Relative paths in diograph.json (+ fix snapshot test)

jvalanen:
TODO:
* [ ] Needs either 
a) feature flag or 
b) reverting `development-content-room/diograph.json` not to have relative paths (because otherwise the E2E tests will fail...)
* [ ] As a part of the Format paper to think about this in a bigger picture

Why this should be merged:
* [ ] generateDiograph snapshot test would be nice to get enabled

---

This doesn't change any functionality because:
* `getImageUrl` currently returns the original `imageUrl`
* `generateDiographWithRelativeImagePaths` is not used by folder connector

May need some more testing first...but would be nice to get the changes merged to the main codebase...

May need some better consideration how this affects to everything: own paper!

Could this be a good case to test feature flags?

---

Changes:

* New function `getImageUrl()` 
  * Is used everywhere in frontend where we retrieve diory.image data
  * Resolves image paths depending on the `diory.image` property 
    *  Starts with http(s)://
    * Development content from repository: starts with `./public` and hosted in localhost:3300
    * Everything else should be just: `file://${diographFolderPath}${imageUrl}`

* Relative paths changed to dev/test content:
  * GET_ROOM.json (used by localhost:3300)
  * development-content-room/diograph.json (used by browser tests)

* Snapshot test fixed
  * Mock `fs.statSync` to get static created & modified timestamps
  * Relative paths to snapshot
  * Use special `generateDiographWithRelativeImagePaths`

* Special `generateDiographWithRelativeImagePaths` 
  * because this is POC and didn't want to change functionality
  * don't know a good way to convert real image paths with relative path in `diory.image`


jvalanen:
Work continues in [Relative paths design paper](https://paper.dropbox.com/doc/Relative-paths--BIiPI7g3If1fkdaGXa5movpXAg-efKwilxekDZX5YAphHCuH)

PR #96, opened 2021-02-17
DDA-83: Filters

opohjola:
Set filter active and inactive individually
Filter diorys when filter is active
Change filtering criteria when filter is active
- Graph → focus diory and zoom -> diory no more than zoom number of links from focus
- Map → zoom or pan view → diory location in view
- Timeline → zoom or pan view → diory date in view
- Text -> query -> diory text includes query


src/react/features/diograph/hooks.js
```diff
@@ -1,28 +1,36 @@
 import { useStore } from '../../store'
 
+const useDiory = (id) => {
+  const [{ diograph = {} }] = useStore((state) => state.diograph)
+  return diograph[id]
+}
```
jvalanen, line 6: I think this as `retrieveDioryFromDiograph` or `getDioryById` . useDiory doesn't feel optimal name for this. But it can also be that I'm not used to it... (and I don't have any better suggestion!)

src/react/features/filters/map/useGenerateMapFilter.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+        it('returns true', () => {
+          const mapFilter = useGenerateMapFilter()
+          expect(mapFilter({ latitude: 15, longitude: 15 })).toEqual(true)
+        })
+      })
+    })
+
+    describe('given diory do not have latitude', () => {
+      it('returns true', () => {
```
jvalanen, line 67: This should be "returns false"?

jvalanen (approved):
I read this through and as a proof for it left one comment! Although that one test should be also fixed and get pull request to green!

src/react/features/filters/map/useGenerateMapFilter.spec.js
jvalanen: And below is another similar one...

jvalanen: This is still here. And still serves as a proof that I read this :blush:

jvalanen (approved):

PR #100, opened 2021-02-25
DDA-98: Open audio and text files in browser

jvalanen:
* More supported file types for images, videos, audio and text (.gif, .flv, .mp3, .pdf...)
  * Commented-out the ones without browser support...
* Open image, video, audio and text dataobjects in browser (from an ugly link hidden in Update tool 😬 , but it works! 🎉 )
* Add audio, video and text dataobject paths to image attribute (temporary solution to enable opening dataobjects in browser...)

NOTE: When you open a dataobject in browser you don't have any other way to get back to the app than: 1) Open developer tools 2) Execute `window.history.back()` in console 😊 , but it works! 🎉 😂

### DIFF
--- a/electron/generators/diory-generator.js
+++ b/electron/generators/diory-generator.js
@@ -18,6 +18,7 @@ function readFileData(type, filePath) {
       return {
         created: fileData.created,
         modified: fileData.modified,
+        image: pathToFileURL(filePath),
         ...readVideo(filePath),
       }
     case 'audio':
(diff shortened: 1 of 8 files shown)


electron/generators/diory-generator.js
```diff
@@ -17,8 +18,15 @@ function readFileData(type, filePath) {
       return {
         created: fileData.created,
         modified: fileData.modified,
+        image: pathToFileURL(filePath),
```
opohjola, line 21: There must be a limit for "just get it working". If we allow implementations like this will our code be a spaghetti sooner than we notice...? This is not even a POC.

If we want to have this, at least we should not use image attribute for something that is not image... 
maybe
 url: pathToFileURL(filePath),

or try out data with schema.org - json-ld e.g.
data: [
     {
            "@context": "https://schema.org",
            "@type": "MusicRecording",
            "audio": "foo-fighters-rope-play.html",
    }
]

PR #101, opened 2021-02-26
DDA-98: Fix map and timeline

jvalanen:


electron/channels/channel-util.js
```diff
@@ -16,7 +16,11 @@ if (process.env.NODE_ENV === 'test') {
 exports.eventHandlerWrapper = function eventHandlerWrapper(channelName, eventHandler) {
   function specificEventHandler(event, params) {
     function success(responseObject) {
-      backendLogger.info('Backend IPC response:', channelName, responseObject)
+      backendLogger.info(
+        'Backend IPC response:',
+        channelName,
+        channelName === 'GET_ROOM' && responseObject ? responseObject.rootId : responseObject
```
opohjola, line 22: How about channel specific logging...?

src/react/features/lenses/timeline/hooks/getTimelineData.spec.js
```diff
@@ -0,0 +1,29 @@
+import { getTimelineData } from './getTimelineData'
+
+const zeroLocation = { lat: 0, lng: 0 }
+const dioryWithUndefinedDate = { date: undefined }
+
+describe('getTimelineData', () => {
```
opohjola, line 6: As a reader of this spec, I would like to read the spec 
- in plain english
- from top to bottom
- as little as possible

Therefore, I suggest that 
- we write the spec in plain English in "describe" and "it"
- we do not set variables in the beginning of the spec as one have to scroll to top to see what e.g. zeroLocations means
- e.g. diory: { date: undefined } already says to "diory with undefined date" -> shorter, no extra variables, less lines, more exact, no scrolling :)

src/react/features/lenses/timeline/hooks/getTimelineData.js
```diff
@@ -41,9 +41,9 @@ const getDioryTimelineData = ({ diory = {}, diorys = [], parent }) => {
   const { lng } = getDioryDateLongitude({ diory, diorys, parent })
   return {
     id: diory.id,
-    center: lng && {
```
opohjola, line 44: useTimelineBounds has default location that requires center to be undefined if there is no location.
  DEFAULT_LOCATION: {
    lat: 0,
    lng: getDateLongitude({ date: new Date() }),
  },
However, this has a bug. It should have new Date().toISOString().
Maybe that fixes this problem.

src/react/features/lenses/timeline/hooks/getTimelineData.js
```diff
@@ -4,10 +4,10 @@ const getAverage = (array = []) =>
 const concat = (array = [], item) => (typeof item !== 'undefined' ? array.concat(item) : array)
 
 export const getDateLongitude = ({ date }) =>
-  (Date.parse(date) / Date.parse('1971-01-01T00:00:00.000Z')) * 10
+  date ? (Date.parse(date) / Date.parse('1971-01-01T00:00:00.000Z')) * 10 : 0
```
opohjola, line 7: I'm afraid this cannot be like this and will cause many problems in timeline. At least it cannot be zero but undefined or null.

opohjola (commented):
Check if useTimelineBounds fixes the timeline problem

PR #110, opened 2021-03-18
DDA-103: Delete tool improvements

jvalanen:
Main improvements:
* Removes also reverse links of the deleted diory
* Confirmation dialog for delete tool
  * Lists all diories and links which are going to be deleted

Implementation related stuff & tests:
* DeleteView: Refactored to use same kind of architecture than Update/CreateTool with View component
* Storybook with different scenarios: diory without links, link without diory and diory with links
* Feature tests which check that diories and links are removed from the Store (using exposing Store contents to Cypress which was introduced in https://github.com/DioryMe/diory-browser-electron/pull/105)
* Unit tests: 
  * deleteDiory and deleteLink reducers
  * useDioryAndLinks hook (which helps to compose the DeleteView)

Design paper: [Delete tool](https://paper.dropbox.com/doc/Delete-tool--BHIdrIy1gRmUDGHIb7pjjHW_Ag-Zsq0CmFTwdvcVb2TQTUam)

### DIFF
--- a/src/react/features/diograph/actions.js
+++ b/src/react/features/diograph/actions.js
@@ -78,13 +78,11 @@ export const deleteLink = (fromDiory, toDiory) => ({
 
 export const deleteLinks = (deletedLinks) => ({
   type: DELETE_LINKS,
-  payload: deletedLinks.map(({ fromDiory, toDiory }) => {
-    return {
-      fromDiory: {
-        ...fromDiory,
-        modified: new Date().toISOString(),
-      },
-      toDiory,
-    }
-  }),
+  payload: deletedLinks.map(({ fromDiory, toDiory }) => ({
+    fromDiory: {
+      ...fromDiory,
+      modified: new Date().toISOString(),
+    },
+    toDiory,
+  })),
 })
--- a/src/react/features/diograph/actions.js
+++ b/src/react/features/diograph/actions.js
@@ -62,13 +62,13 @@ export const createLink = (diory, link) => ({
   },
 })
 
-export const deleteLink = (diory, link) => ({
+export const deleteLink = (diory, linkedDiory) => ({
   type: DELETE_LINK,
   payload: {
     diory: {
       ...diory,
       modified: new Date().toISOString(),
     },
-    link,
+    linkedDiory,
   },
 })
--- a/src/react/features/diograph/actions.js
+++ b/src/react/features/diograph/actions.js
@@ -6,6 +6,7 @@ import {
   DELETE_DIORY,
   CREATE_LINK,
   DELETE_LINK,
+  DELETE_LINKS
 } from './actionsTypes'
 
 export const getRoom = (room) => ({
@@ -62,13 +63,26 @@ export const createLink = (diory, link) => ({
   },
 })
 
-export const deleteLink = (diory, linkedDiory) => ({
+export const deleteLink = (fromDiory, toDiory) => ({
   type: DELETE_LINK,
   payload: {
-    diory: {
-      ...diory,
+    fromDiory: {
+      ...fromDiory,
       modified: new Date().toISOString(),
     },
-    linkedDiory,
+    toDiory,
   },
+})
+
+export const deleteLinks = (deletedLinks) => ({
+  type: DELETE_LINKS,
+  payload: deletedLinks.map(({ fromDiory, toDiory }) => {
+    return {
+      fromDiory: {
+        ...fromDiory,
+        modified: new Date().toISOString(),
+      },
+      toDiory
+    }
+  })
 })
--- a/src/react/features/diograph/actions.js
+++ b/src/react/features/diograph/actions.js
@@ -6,7 +6,7 @@ import {
   DELETE_DIORY,
   CREATE_LINK,
   DELETE_LINK,
-  DELETE_LINKS
+  DELETE_LINKS,
 } from './actionsTypes'
 
 export const getRoom = (room) => ({
@@ -63,6 +63,8 @@ export const createLink = (diory, link) => ({
   },
 })
 
+// Currently this is not used anywhere
+// - same as deleteLinks but takes {} as an argument instead of [{}]
 export const deleteLink = (fromDiory, toDiory) => ({
   type: DELETE_LINK,
   payload: {
@@ -82,7 +84,7 @@ export const deleteLinks = (deletedLinks) => ({
         ...fromDiory,
         modified: new Date().toISOString(),
       },
-      toDiory
+      toDiory,
     }
-  })
+  }),
 })
--- a/src/react/features/tools/delete/DeleteTool.js
+++ b/src/react/features/tools/delete/DeleteTool.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import { useStore } from '../../../store'
 import { useLinkDiory, useFocus } from '../../diograph/hooks'
-import { useDeleteTool } from './useDeleteDioryAndLinks'
+import { useDeleteView } from './useDeleteView'
 
 import DeleteView from './DeleteView'
 
@@ -12,7 +12,7 @@ const DeleteTool = () => {
   const [{ link }] = useStore((state) => state.navigation)
   const { diory: focusDiory } = useFocus()
   const { diory: linkDiory } = useLinkDiory()
-  const props = useDeleteTool(focusDiory, linkDiory)
+  const props = useDeleteView(focusDiory, linkDiory)
 
   if (DELETE_TOOL_BUTTON === active && !!link) {
     return <DeleteView {...props} />
--- a/src/react/features/tools/delete/DeleteTool.js
+++ b/src/react/features/tools/delete/DeleteTool.js
@@ -13,7 +13,7 @@ const DeleteTool = () => {
   const { diory } = useFocus()
 
   return DELETE_TOOL_BUTTON === active && !!link ? (
-    <DeleteView diory={linkDiory} focus={diory} />
+    <DeleteView focus={diory} linkDiory={linkDiory} />
   ) : null
 }
 
--- a/src/react/features/tools/delete/DeleteTool.js
+++ b/src/react/features/tools/delete/DeleteTool.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import { useStore } from '../../../store'
 import { useLinkDiory, useFocus } from '../../diograph/hooks'
-import { useDeleteDioryAndLinks, useGenerateDeletedLinks } from './useDeleteDioryAndLinks'
+import { useDeleteTool } from './useDeleteDioryAndLinks'
 
 import DeleteView from './DeleteView'
 
@@ -12,14 +12,7 @@ const DeleteTool = () => {
   const [{ link }] = useStore((state) => state.navigation)
   const { diory: focusDiory } = useFocus()
   const { diory: linkDiory } = useLinkDiory()
-
-  const isFocusDeleted = focusDiory && linkDiory ? focusDiory.id === linkDiory.id : false
-  const deletedDiory = isFocusDeleted ? focusDiory : null
-
-  const links = focusDiory && linkDiory ? (isFocusDeleted ? focusDiory.links || {} : { [linkDiory.id]: { id: linkDiory.id } }) : []
-  const deletedLinks = useGenerateDeletedLinks(focusDiory, links, isFocusDeleted)
-
-  const props = useDeleteDioryAndLinks(deletedDiory, deletedLinks)
+  const props = useDeleteTool(focusDiory, linkDiory)
 
   if (DELETE_TOOL_BUTTON === active && !!link) {
     return <DeleteView {...props} />
--- a/src/react/features/tools/delete/DeleteTool.js
+++ b/src/react/features/tools/delete/DeleteTool.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import { useStore } from '../../../store'
 import { useLinkDiory, useFocus } from '../../diograph/hooks'
-import { useDeleteDioryAndLinks } from './useDeleteDioryAndLinks'
+import { useDeleteDioryAndLinks, useGenerateDeletedLinks } from './useDeleteDioryAndLinks'
 
 import DeleteView from './DeleteView'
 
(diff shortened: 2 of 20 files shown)


src/react/features/diograph/actions.js
```diff
@@ -62,13 +62,13 @@ export const createLink = (diory, link) => ({
   },
 })
 
-export const deleteLink = (diory, link) => ({
+export const deleteLink = (diory, linkedDiory) => ({
```
opohjola: I think it should be link. Link is just [key]: { id }, it is not the same as linked diory.

src/react/features/tools/delete/DeleteTool.js
```diff
@@ -0,0 +1,20 @@
...
+import { useLinkDiory, useFocus } from '../../diograph/hooks'
+
+import DeleteView from './DeleteView'
+
+import { DELETE_TOOL_BUTTON } from './buttons'
+
+const DeleteTool = () => {
+  const [{ active }] = useStore((state) => state.buttons)
+  const [{ link }] = useStore((state) => state.navigation)
+  const linkDiory = useLinkDiory().diory
```
opohjola: const { diory: linkDiory } =

src/react/features/tools/delete/DeleteTool.js
```diff
@@ -0,0 +1,20 @@
...
+  const { diory } = useFocus()
```
opohjola: const { diory: focusDiory } for consistency

src/react/features/tools/delete/useDeleteDioryAndLinks.js
```diff
@@ -0,0 +1,66 @@
+import { useStore, useDispatch } from '../../../store'
+import { setOpen, setInactive } from '../../buttons/actions'
+import { goBackward, setSelectedLink } from '../../navigation/actions'
+import { deleteDiory, deleteLink } from '../../diograph/actions'
+
+const generateDeletedLinks = (focus, links, isFocusDeleted) => {
```
opohjola: use -prefix as hook

src/react/features/tools/delete/useDeleteDioryAndLinks.js
```diff
@@ -0,0 +1,66 @@
...
+  const dispatch = useDispatch()
+  return {
+    deletedDiory,
+    deletedLinks,
+    deleteDioryAndLinks: () => {
+      deletedLinks.forEach((deletedLink) => {
+        if (deletedLink.fromDiory.id === deletedLink.toDiory.id) {
+          return
+        }
+        dispatch(deleteLink(deletedLink.fromDiory, deletedLink.toDiory))
```
opohjola: instead of looping one action, create new action: deleteLinks(deletedLinks)

src/react/features/tools/delete/useDeleteDioryAndLinks.js
```diff
@@ -0,0 +1,66 @@
...
+          : null
+      })
+      .filter(Boolean)
+    return linkedDiories.concat(reverseLinkedDiories)
+  }
+
+  return linkedDiories
+}
+
+export const useDeleteDioryAndLinks = (focus, clickedDiory) => {
```
opohjola: if there is no reason for props, use hooks

src/react/features/tools/delete/DeleteView.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Modal from '../../../components/Modal'
+import { useDeleteDioryAndLinks } from './useDeleteDioryAndLinks'
+
+const DeleteView = ({ focus, clickedDiory }) => {
```
opohjola: Is there a reason for some data coming as props and some from hooks?
- maybe all from hooks or all as props?

opohjola: If there is one spec for useDeleteDioryAndLinks and a story for the DeleteView, then I would prefer all data as props.

opohjola (commented):
Some changes proposed

src/react/features/tools/delete/useDeleteDioryAndLinks.js
```diff
@@ -0,0 +1,66 @@
...
+  const isFocusDeleted = focus.id === clickedDiory.id
+  const deletedDiory = isFocusDeleted ? focus : null
+  const links = isFocusDeleted ? focus.links || {} : { [clickedDiory.id]: { id: clickedDiory.id } }
+  const deletedLinks = generateDeletedLinks(focus, links, isFocusDeleted)
```
opohjola: One of redux style guides is:
"Put as Much Logic as Possible in Reducers#
Wherever possible, try to put as much of the logic for calculating a new state into the appropriate reducer, rather than in the code that prepares and dispatches the action"

According to that style guide, the most this logic should be in a reducer.

src/react/features/diograph/actions.js
jvalanen: But it's a diory that is given to it: https://github.com/DioryMe/diory-browser-electron/blob/develop/src/react/features/tools/delete/useDeleteTool.js#L18

That's why I made the change to make it explicit. And now I'm confused why you say it should be a link 😳

src/react/features/tools/delete/DeleteTool.js
jvalanen: Oh, that's a new syntax for me! 

But do we really want to use it systematically everywhere? Every time when we don't use the key name as a variable name (but want to rename it to something else)

If so, then we should change our linting rules and use: 
```
    "prefer-destructuring": ["error", {
      "array": false,
      "object": true
    }, {
      "enforceForRenamedProperties": true
    }]
```
https://eslint.org/docs/rules/prefer-destructuring

But then we will get a lot of errors in other places of the code:
https://github.com/DioryMe/diory-browser-electron/runs/2164703431?check_suite_focus=true

Alternative is to use it with `enforceForRenamedProperties: false` where this particular case `const linkDiory = useLinkDiory().diory` is OK and they can be used with "own consideration". But in the future all the `const diory = useLinkDiory().diory` would give a linting error and needs to be corrected to `const { diory } = useLinkDiory()`

Which do you prefer?

jvalanen: I think everytime we need to make this kind of comment in review we should think about if we should change our linting rules and make a suggestion.

Needing to pay attention for this kind of syntax-details makes it difficult to pay attention to the more valuable things such as functionality, testing, naming etc.

src/react/features/tools/delete/DeleteTool.js
jvalanen: This is the same case as the previous comment

1) Is it ok to do `const focusDiory = useFocus().diory` ? 
2) Or use that a bit fancy `{ diory: focusDiory }` syntax

src/react/features/tools/delete/DeleteView.js
jvalanen: This is now refactored.
1) `useDeleteDioryAndLinks` renamed to `useDeleteView` and extracted out from `DeleteView` and is now called in `DeleteTool`
2) `<DeleteView />` gets all its props from `const props = useDeleteView`
3) Storybook gives just `deletedDiory` or `deletedLinks` (+ actions) straight as props and doesn't need `<StoryProvider />` anymore! 🎉

src/react/features/tools/delete/useDeleteDioryAndLinks.js
jvalanen: This is fixed.

src/react/features/tools/delete/useDeleteDioryAndLinks.js
jvalanen: I will try this tomorrow as I also make a new suggestion from https://github.com/DioryMe/diory-browser-electron/pull/111 . Need to do this kind of stuff there too...

jvalanen (commented):
I refactored `<DeleteView {...props} />` with `useDeleteView` if you could check that I understood correctly and it looks better now. Now there's now logic / useStore inside of DeleteView component anymore, it is just for rendering the dialog and setting actions for Delete and Cancel/X buttons.

Tomorrow I'll take a look at the reducer stuff if I those changes would make this better.

Having good tests make this kind of refactoring kind of trivial as the functionality is verified by the tests :)

src/react/features/tools/delete/useDeleteDioryAndLinks.js
jvalanen: Good suggestion! This is now done and made the code much better!

src/react/features/tools/delete/useDeleteDioryAndLinks.js
jvalanen: Now all the data is given as props and no useStore inside the `<DeleteView>` component

src/react/features/tools/delete/useDeleteDioryAndLinks.js
jvalanen: Thanks for the style guide quote, didn't know that!

I refactored the whole thing and it came out so much clearer! (although also took a good hour to do...)
Didn't end up putting the logic into the reducer because our DeleteView needs the list of deleted links too. Otherwise it would be a good idea (and also pretty trivial task after the refactor...) to put the logic there. And maybe if we need to delete stuff elsewhere we have to consider make own helpers for the deleting stuff...

(in fact, don't we currently already have a delete tool on map... 🤔 )

src/react/features/diograph/actions.js
opohjola: yes, my mistake. It does not have key, but just linked diory.

src/react/features/tools/delete/DeleteTool.js
opohjola: Good idea about considering linting rules in this kind of cases!
I think we cannot enforce this particular rule as many cases e.g. in tests it makes things less clear. 

enforceForRenamedProperties: false sounds good. 

Is in this case const { linkDiory } = useLinkDiory() actually the best option? :)

src/react/features/tools/delete/DeleteTool.js
opohjola: { diory: focusDiory } is good in that sense that it clearly communicates that this variable is renamed.
Also, as destructuring the generally the prefeffed way, I would prefer that also when one needs to rename.

src/react/features/tools/delete/DeleteView.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Modal from '../../../components/Modal'
+
+const DeleteView = ({ deleteDioryAndLinks, resetView, deletedDiory, deletedLinks }) => (
```
opohjola: This is generally ok, however on thing to consider:

I have typically made Views as stupid components as possible and put the most of the logic to hooks. This is especially case when one component is used for several purposes, which this is not now but who knows will be used tomorrow.

Here this would mean e.g. following prop names ({ onDone, onCancel, diory, diorys })
onDone, onCancel: because View do not need to know how it is used.
diory, diorys: the props names used generally throughout the app for indication one and several diorys.

useDeleteView would have the logic deleted linked diory texts:
text: &quot;
            {deletedLink.fromDiory.text || deletedLink.fromDiory.id}
            &quot; -&gt; &quot;
            {deletedLink.toDiory.text || deletedLink.toDiory.id}
            &quot;


This is now one coding style proposal. What we do with these? Collect to Coding Style paper under title Proposal. I'll do that.

opohjola: in Best Practises

src/react/features/tools/delete/useDeleteView.spec.js
```diff
@@ -0,0 +1,99 @@
...
+      expect(mockDispatch).toHaveBeenCalledTimes(6)
+      expect(mockDispatch).toHaveBeenCalledWith(deleteDiory(diory))
+      expect(mockDispatch).toHaveBeenCalledWith(deleteLinks(expectedDeleteLinksArguments))
+      expect(mockDispatch).toHaveBeenCalledWith(goBackward())
+      expect(mockDispatch).toHaveBeenCalledWith(setInactive())
+      expect(mockDispatch).toHaveBeenCalledWith(setSelectedLink())
+      expect(mockDispatch).toHaveBeenCalledWith(setOpen(false))
+    })
+
+    it('delete diory in focus without links', () => {
```
opohjola: Assign
---empty line---
Act
---empty line---
Assert

structure makes it easier to read tests (once you know it is used :)

src/react/features/tools/delete/useDeleteView.js
```diff
@@ -0,0 +1,64 @@
+import { useStore, useDispatch } from '../../../store'
+import { setOpen, setInactive } from '../../buttons/actions'
+import { goBackward, setSelectedLink } from '../../navigation/actions'
+import { deleteDiory, deleteLinks } from '../../diograph/actions'
+
+const useLinkedDiories = (focusDiory, diograph) =>
```
opohjola: FIX: use -prefix in non-hook

src/react/features/tools/delete/DeleteTool.js
```diff
@@ -0,0 +1,24 @@
...
+import { useDeleteView } from './useDeleteView'
+
+import DeleteView from './DeleteView'
+
+import { DELETE_TOOL_BUTTON } from './buttons'
+
+const DeleteTool = () => {
+  const [{ active }] = useStore((state) => state.buttons)
+  const [{ link }] = useStore((state) => state.navigation)
+  const { diory: focusDiory } = useFocus()
```
opohjola: RECOMMENDATION: Move useFocus and useLinkDiory to useDeleteView

opohjola (approved):

src/react/features/tools/delete/useDeleteView.spec.js
jvalanen: I don't understand how it is not the case in here: 
1. Assign is in the beforeEach, 
2. Act is at first and 
3. Assert is after an empty line

🤔

jvalanen: But this has been completely rewritten afterwards so it doesn't realy matter...

PR #111, opened 2021-03-21
DDA-000: Refactor buttons reducers

jvalanen:
We don't have good (enough?) test coverage for this kind of refactorings but I wanted to make this suggestion as PR anyway.

* Rename buttons reducers with SET_ -prefix (openButtons, activateButtons, inactivateButtons)
* Improve naming: `buttons` -> `addedButtons`/`removedButtons`
* Improve destructuring: `{ payload: { diory } }` & `diory` instead of `{ payload }` & `payload.diory`


src/react/features/buttons/actions.js
```diff
@@ -1,25 +1,21 @@
-import { ADD_BUTTONS, REMOVE_BUTTONS, SET_OPEN, SET_ACTIVE, SET_INACTIVE } from './actionsTypes'
+import { ADD_BUTTONS, REMOVE_BUTTONS, SET_OPEN, SET_ACTIVE } from './actionsTypes'
 
-export const addButtons = (buttons) => ({
+export const addButtons = (addedButtons) => ({
```
opohjola, line 9: For me this is unnecessary repetition: addButton(buttons) already says that those buttons are addedButtons.

src/react/features/buttons/actions.js
```diff
@@ -1,25 +1,21 @@
...
   payload: { open },
 })
 
-export const setActive = (active) => ({
+export const setActive = (buttonId) => ({
   type: SET_ACTIVE,
-  payload: { active },
-})
-
-export const setInactive = () => ({
```
opohjola, line 23: One reduc style guide:
"However, we recommend trying to treat actions more as "describing events that occurred", rather than "setters". Treating actions as "events" generally leads to more meaningful action names, fewer total actions being dispatched, and a more meaningful action log history."

According to this style guide, we should go to other direction and change setOpen to openButtons and add closeButtons; and rename setActive to activeButton and inactivateButton.

opohjola (approved):

src/react/features/buttons/actions.js
jvalanen: So you'd prefer `buttons` instead of `addedButtons`? That is more clear / cleaner for your eye?

Other reason to change this name was to get rid of `payload.` in the reducers and make those also cleaner (at least for me...). `buttons` is already used there as a part of `omit` stuff so there's a need for renaming if we want to remove `payload.` and I thought this `removedButtons` is a good name and even makes it even more clear which buttons those are...
image

That "cleaning" has been done in several other places as well.
Do you think that is also unnecessary?

src/react/features/buttons/actions.js
jvalanen: Thanks for the good quote from style guide and good suggestion for renaming!

Renaming done: 
```
setOpen -> openButtons
setActive -> activateButton
setInactive -> inactivateButton
```

closeButtons wasn't added because it is not used anywhere... Everytime we close buttons we want also to inactivateButton so it doesn't make sense to add it if it is not used...

jvalanen (commented):
Thanks for the comments, changes made! 

I'm pretty happy with this refactoring, it makes our codebase a little bit easier to understand and enforces the conventions!

src/react/features/buttons/actions.js
opohjola: just giving my opinion... my rule here is: no extra / unnecessary text.

if needed, renaming can be used in reducers { buttons: removedButtons }

in reducers, I have not destructured payload and state, as then it communicates consistently which prop is coming from state and which from payload, but "removedButtons" does the same so that's ok.

opohjola (approved):

PR #125, opened 2021-04-15
DDA-118: Resolve mime type instead of detecting file extensions

jvalanen:
* Generate diory data in two parts: 
  * `baseData` with created, modified and text which are composed from basic file attributes (name, birthtime, mtime)
  * `typeSpecificData` with date, geo, image etc. which are composed from file metadata
* Add [file-type package](https://github.com/sindresorhus/file-type) for MIME type detection
* Remove "file extension detection logic" and use MIME type instead of it
* Merge parts of file-reader into diory-generator (and remove the rest)
* Remove video-reader 
* Add TODO/FIXME comments to remove video attribute


electron/generators/diory-generator.spec.js
```diff
@@ -1,188 +0,0 @@
-const { v4: uuid } = require('uuid')
```
jvalanen, line 1: What do you think? Should we have unit test for generateDioryFromFile? What kind of?

electron/generators/diory-generator.spec.js
```diff
@@ -1,188 +0,0 @@
...
-
-          const diory = await act()
-
-          expect(diory[prop]).toEqual(`some-file-${prop}`)
-        })
-      })
-    })
-  })
-
-  describe('generateDioryFromFolder', () => {
```
jvalanen: This should have not been removed!!
+ include extension in text attribute!

PR #138, opened 2021-05-06
DDA-136: Don't interpolate anything

jvalanen:
* Show only diories with dates on timeline
* Show only diories with locations on map
* Don't show Filters-diory on timeline or map
  * First linked diory shown instead
* Default bounds for timeline when there's only 0 or 1 diories with dates: shows 1990-2030

### DIFF
--- a/src/react/features/filters/useFilteredDiorys.js
+++ b/src/react/features/filters/useFilteredDiorys.js
@@ -31,9 +31,7 @@ export const useFilteredDiorys = () => {
   const diorys = useDiorys(dioryIds)
   return useMemo(
     () => ({
-      diory: {
-        text: 'Filter',
-      },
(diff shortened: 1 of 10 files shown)


src/react/features/filters/useFilteredDiorys.js
```diff
@@ -31,9 +31,7 @@ export const useFilteredDiorys = () => {
   const diorys = useDiorys(dioryIds)
   return useMemo(
     () => ({
-      diory: {
-        text: 'Filter',
-      },
+      diory: diorys[0],
```
jvalanen, line 34: Now when we don't interpolate geo/date to focus diory, we can't give `{ text: 'Filter' }` diory to the map lens.

So quick fix is to have "first linked diory" in focus instead.

PR #182, opened 2021-11-15
Visual regression compare tool v0.1

jvalanen:


opohjola (approved):
Test and production code could be separated better at some point.

jvalanen:
Splitted in two: #207 and #205

PR #225, opened 2022-02-09
WIP: DDA-224: Open website in browser

jvalanen:
WIP:
* [ ] <ScrollBackground> in GridView.js most probably is not the right place for the click
* [ ] Update tool, memory grid clicks etc. triggers this unintentionally (because I don't know how to manage that bubbling thing...)

* Now there can be two dataobject sources for website: contentUrl from OPEN_TOOL and url from this background click!


jvalanen:
Replaced by https://github.com/DioryMe/diory-browser-electron/pull/235

PR #230, opened 2022-02-21
POC: Delete with dataobject

jvalanen:
NOTE: CI builds fail because DioryMe/diograph-js repository doesn't have /dist folder a.k.a. Javascript compiled from Typescript => what would be a good convention for this? Should /dist folder be included to the repo? Should CI go to ./diograph-js folder and run `yarn build` to create the /dist folder?

Example implementation of "Delete with dataobjects"
* Works
* DeleteDiory removes dataobject always if available


jvalanen:
Part of POC-2022-02

PR #231, opened 2022-02-21
POC-000: Thumbnails

jvalanen:
Post-generates thumbnails for imported folder
* Works
* Horrible code and kind of proves that thumbnails shouldn't be done without Room & DiographJson
* Puts thumbnails into same folder as imported folder (no explicit /images folder)


jvalanen:
Part of POC

PR #243, opened 2022-03-11
DDA-242: Prevent unintentional key press on focus

opohjola:


src/react/components/Button.js
```diff
@@ -13,7 +13,6 @@ const Button = ({ style = {}, data, active, onClick }) => (
   <div
     role="button"
     onClick={onClick}
-    onKeyDown={onClick}
```
jvalanen, line 16: What was purpose of this onKeyDown in the first place? Do we lose some feature now as it is removed?

opohjola: eslint proposed it, if I remember right.

jvalanen: And that fixed the "Unintentional key press on focus" issue?
One line fix? 😳 👌

jvalanen (approved):

PR #252, opened 2022-06-08
WIP: "List content source tool 0.0.1"

jvalanen:
The Original GridLens2 feat. new diograph-js which:
* demo for listContentSource tool a.k.a. localDiographGenerator
* creates connection to user's local disk as a "content source"
* lets user to browse the "content source" by generating diories from the contents
* saves room.json to user's local disk and saves it every time more diories are retrieved to connection's diograph

This kind of thing should work also with other kind of "content sources", not just folder structure: Whatsapp archive, .mbox etc.


jvalanen:
Too long time since last commit and nobody's doing anything... Doesn't make sense to merge anymore.

PR #255, opened 2022-06-24
WIP: DDA-254: GridLens2 in ImportFolderTools modal

jvalanen:
Applied version from POC-251-list-content-source (#252) where new kind of "dummied GridLens" is rendered inside of the ImportFolderTools modal.

Done in a "feature flag" way where (almost) nothing old is modified but all the components that are used are duplicated so it's easy and safe to start using the old.

* diograph-js & @diograph/file-generator packages are in use
* new diograph-js ElectronServer backend is added to preload.js


jvalanen:
Too long time since last commit and nobody's doing anything... Doesn't make sense to merge anymore.

PR #260, opened 2024-02-11
DDA-259: Create rooms sidebar and diograph graph

opohjola:
Diosphere sidebar
- rooms navigation
Navigation
Graph lens
Diograph with '/' root


.gitmodules
```diff
@@ -1,3 +1,9 @@
 [submodule "diograph-js"]
 	path = diograph-js
 	url = git@github.com:DioryMe/diograph-js.git
+[submodule "@diograph/file-generator"]
+	path = @diograph/file-generator
+	url = git@github.com:DioryMe/file-generator.git
+[submodule "@diograph/folder-generator"]
+	path = @diograph/folder-generator
+	url = https://github.com/DioryMe/folder-generator.git
```
jvalanen: All these would be better be used as npm packages and added to package.json

Current versions:
- file-generator 0.1.2-rc4
- folder-generator 0.2.1-rc3
- @diograph/diograph 0.2.2-rc2

So these submodules could be removed.
No need to do this for this PR but keep in mind in the future

electron/folder-connector/index.js
```diff
@@ -0,0 +1,10 @@
+const { contextBridge } = require('electron')
+const { readJson } = require('./readJson')
+const { saveJson } = require('./saveJson')
+
+contextBridge.exposeInMainWorld('folderConnector', {
+  getDiograph: readJson('diograph.json'),
+  saveDiograph: saveJson('diograph.json'),
+  getDiosphere: readJson('diosphere.json'),
+  saveDiosphere: saveJson('diosphere.json'),
```
jvalanen, line 9: In the future these should use @diograph/diograph commands like done in diograph-cli:
1. Construct RoomClient with address and client
2. Construct room with client
3. Load room
4. Use room commands: 
  - `room.diograph.toObject()` or `room.diograph.queryDiograph()`
  - `room.diograph.addDiory({ ... })` and  `room.saveRoom()`

https://github.com/DioryMe/diograph-cli/blob/commander/src/configManager.ts#L84

public/diory-demo-content/diograph.json
```diff
@@ -3,7 +3,7 @@
   "diograph": {
```
jvalanen, line 3: This seems to be old format of diograph... Shouldn't be `rootId` there...

Using demo-content-room repository also here would be preferable.

public/diory/diosphere.json
```diff
@@ -0,0 +1,81 @@
+{
```
jvalanen, line 1: Interesting, now I remember us talking about this diosphere.json!

In diograph-cli there's a config file called `.dcli` used for this purpose, it looks like this:
```
ffmpegPath=/opt/homebrew/bin/ffmpeg

[focus]
connectionInFocus=/tmp
roomInFocus=/tmp

[rooms./tmp]
address=/tmp
roomClientType=LocalClient
```

I wonder where this should be saved? From user's perspective who hosts this: diory-browser-electron app or the same file-storage host where also rooms are hosted?

public/diory/diosphere.json
```diff
@@ -0,0 +1,81 @@
...
+        }
+      ],
+      "doors": [
+        {
+          "id": "first-image-room-id"
+        },
+        {
+          "id": "second-image-room-id"
+        }
+      ]
```
jvalanen, line 41: I think this needs also type. Is it Local, S3, IPFS or which? Client needs that information...

electron/preload.js
```diff
@@ -4,10 +4,10 @@ const { fileURLToPath } = require('url')
 const { channels } = require('../src/shared/constants')
 
 const { importFolder } = require('./lib/import-folder')
-const { getDiograph } = require('./lib/get-diograph')
 const { getDioryFolderLocation } = require('./lib/get-diory-folder-location')
 const { setDioryFolderLocation } = require('./lib/set-diory-folder-location')
-const { saveDiograph } = require('./lib/save-diograph')
+
+require('./folder-connector')
```
jvalanen, line 10: Electron 28.0.0 finally supports ES modules: https://www.electronjs.org/docs/latest/tutorial/esm

Need to upgrade to that at some point!

jvalanen (commented):
Left some random comments :D 

I really need this current version in action. Seems that a lot of has changed!

We also need to get on the same page how we see the diosphere. room, diograph, folder-generator, file-generator, folder-connector etc. stuff!

I would really love to get @diograph/diograph etc. npm packages to work with this Electron thingie!!

opohjola:
I'll merge this to develop even if Cypress and AppVeyor fails.
Let's fix those once there are a usable version.

